PR #2, opened 2020-03-25
Added machine learning efficacy

k15z:


sdmetrics/efficacy/base.py
```diff
@@ -0,0 +1,99 @@
...
+    name = ""
+    target_table = ""
+    target_column = ""
+
+    metric_unit = ""
+    metric_goal = Goal.IGNORE
+    metric_domain = (float("-inf"), float("inf"))
+
+    def fit(self, X, y):
+        """This function implements a fit procedure which trains a binary
```
kveerama, line 18: This doc string seems to be from adversarial metrics. Needs an update to MLefficacy

sdmetrics/efficacy/base.py
```diff
@@ -0,0 +1,99 @@
+import numpy as np
+from sklearn.model_selection import KFold
+
+from sdmetrics import Goal, Metric
+
+
+class MLEfficacy():
+
+    name = ""
+    target_table = ""
```
kveerama, line 10: target_table_name, target_column_name ??

sdmetrics/efficacy/base.py
```diff
@@ -0,0 +1,99 @@
...
+                "table:%s" % self.target_table,
+                "column:%s" % self.target_column,
+            ]),
+            goal=self.metric_goal,
+            unit="delta_%s" % self.metric_unit,
+            domain=(-domain, domain),
+            description="Diff in score on real when trained on synthetic vs real."
+        )
+
+    def _cross_val_score(self, real_table, synthetic_table):
```
kveerama, line 76: we should change this to evaluate_score, since this is not a cross validation score in the traditional sense. Also, a doc string to explain the process.

sdmetrics/efficacy/base.py
```diff
@@ -0,0 +1,99 @@
...
+                "column:%s" % self.target_column,
+            ]),
+            goal=self.metric_goal,
+            unit=self.metric_unit,
+            domain=self.metric_domain,
+            description="Score on the real test set when trained on synthetic data."
+        )
+
+        # Score on synthetic minus score on real. Evaluated on real.
+        domain = self.metric_domain[1] - self.metric_domain[0]
```
kveerama, line 61: I think this variable should be named delta_domain since this seems to be the domain of the delta_score, and not the domain of the original metric.

sdmetrics/efficacy/base.py
```diff
@@ -0,0 +1,99 @@
...
+
+        synthetic_X = synthetic_table.loc[:, synthetic_table.columns != self.target_column].values
+        synthetic_y = synthetic_table[self.target_column].values
+
+        delta_scores = []
+        synthetic_scores = []
+        kf = KFold(n_splits=3, shuffle=True)
+
+        for train_index, test_index in kf.split(real_X, real_y):
+            # Train a model on the real dataset.
```
kveerama, line 88: Train a model on the real dataset and test on the real dataset.

sdmetrics/efficacy/base.py
```diff
@@ -0,0 +1,99 @@
...
+            self.fit(real_X[train_index], real_y[train_index])
+            real_score = self.score(real_X[test_index], real_y[test_index])
+
+            # Train a model on the synthetic dataset.
```
kveerama, line 92: Train a model on the synthetic dataset and test it on the real test dataset.

PR #7, opened 2020-06-26
Minor bugfixes

csala:
* Also remove the child key from the tables before running parent_child_detection metrics
* Handle an edge-case in cstest when there is only one category

### DIFF
--- a/sdmetrics/__init__.py
+++ b/sdmetrics/__init__.py
@@ -4,7 +4,7 @@
 
 __author__ = 'MIT Data To AI Lab'
 __email__ = 'dailabmit@gmail.com'
-__version__ = '0.0.1.dev2'
+__version__ = '0.0.1.dev3'
 
 from sdmetrics.evaluation import evaluate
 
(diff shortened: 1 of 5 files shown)


k15z:
Just to verify some intuition, here is the expected behavior of chisquared:

```
import numpy as np
from sdmetrics.statistical.univariate import CSTest
```

This will return 1.0 since it's a weird edge case with constant values.

```
real_column = np.array(["no", "no", "no"])
synth_column = np.array(["no", "no", "no"])
CSTest.metric(real_column, synth_column)
```

This will be somewhere between 0.0-1.0 - let's call it q - since it's possible due to random chance that the synthetic data doesn't contain "no".

```
real_column = np.array(["yes", "yes", "no"])
synth_column = np.array(["yes", "yes", "yes"])
CSTest.metric(real_column, synth_column)
```

This will be somewhere between 0.0-1.0 - let's call it `p` - such that `p` < `q` since it's possible due to random chance that the synthetic data doesn't contain "yes" but it's not that likely based on the observed frequencies.

```
real_column = np.array(["yes", "yes", "no"])
synth_column = np.array(["no", "no", "no"])
CSTest.metric(real_column, synth_column)
```

The following will return 0.0 since the synthetic data containing a value that is not in the real data is bad.

```
real_column = np.array(["yes", "yes", "yes"])
synth_column = np.array(["yes", "yes", "no"])
CSTest.metric(real_column, synth_column)
```

PR #23, opened 2020-12-07
Update conda instructions

fealho:
See title.


csala:
Pytorch is not used by SDMetrics, so in this case we can skip adding it to the instructions.

PR #50, opened 2021-02-10
Fixes the sktime issue for conda

fealho:
Updates the dependency requirements in meta.yaml to be able to use the latest versions of sktime, since only v0.5.0 and higher are available in conda-forge. It also makes a small change in the code, since sktime v0.5.0 introduces a breaking API update.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -20,10 +20,10 @@ requirements:
     - pytest-runner
     - scikit-learn >=0.20,<1
     - scipy >=1.1.0,<1.6.0
-    - numpy >=1.15.4,<2
+    - numpy >=1.15.0,<2
     - pandas >=1,<1.1.5
-    - seaborn >=0.9,<0.11
-    - matplotlib >=2.2.2,<3.2.2
+    - seaborn
+    - matplotlib
     - pomegranate >=0.13.0,<0.13.5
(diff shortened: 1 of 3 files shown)


conda/meta.yaml
```diff
@@ -20,25 +20,29 @@ requirements:
...
   run:
     - python
     - scikit-learn >=0.20,<1
     - scipy >=1.1.0,<1.6.0
-    - numpy >=1.15.4,<2
+    - numpy >=1.15.0,<2
     - pandas >=1,<1.1.5
-    - seaborn >=0.9,<0.11
-    - matplotlib >=2.2.2,<3.2.2
+    - seaborn
```
csala, line 39: Why are we removing `seaborn` and `matplotlib` version ranges?

setup.py
```diff
@@ -13,10 +13,10 @@
 
 install_requires = [
     'scikit-learn>=0.20,<1',
-    'sktime>=0.4,<0.5',
+    'sktime>=0.4,<0.6',
     'tsfresh>=0.15,<1',
     'scipy>=1.1.0,<1.6.0',
-    'numpy>=1.15.4,<2',
+    'numpy>=1.15.0,<2',
```
csala, line 19: Why this change from `.4` to `.0`?

PR #55, opened 2021-03-22
Implemented Privacy Metrics

fealho:
Resolves #38.


csala (changes requested):
It seems like the error from Github Actions happens because of the latest torch release. We could try capping it to "torch<1.8" on the workflow yaml file. Also see why here it fails but other projects do not.

Also: `sdmetrics/single_table/README.md` should be updated to add the new metrics.

csala:
I just re-triggered the last build that failed because of the GMLogLikelihood noisy output. I also opened issue #56 to work around this problem.

csala (approved):
It looks great now!

PR #57, opened 2021-03-24
Normalize

### LINKED ISSUE #51: Add normalize method to metrics
### Description

Currently metrics return one score which can be defined in an arbitrary range and can be either a MAXIMIZATION or MINIMIZATION metric.

We could add a `normalize` method to all the base classes (with specific overrides for the individual metrics that require it) which gets the raw score as input and returns it normalized between 0 and 1 and using always MAXIMIZATION goal.

### Usage

```python
raw_score = WhateverMetric.compute(real, synthetic, metadata)
normalized = WhateverMetric.normalize(raw_score)
```

fealho:
Resolve #51.

### DIFF
--- a/sdmetrics/base.py
+++ b/sdmetrics/base.py
@@ -1,5 +1,6 @@
 """BaseMetric class."""
-
+import numpy as np
+from sdmetrics.goal import Goal
 
 class BaseMetric:
     """Base class for all the metrics in SDMetrics.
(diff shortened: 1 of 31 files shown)


sdmetrics/base.py
```diff
@@ -58,3 +61,47 @@ def compute(real_data, synthetic_data):
...
+            score = (raw_score - min_value) / (max_value - min_value)
+
+        elif not is_finite(min_value) and is_finite(max_value):
+            score = np.exp(raw_score - max_value)
+
+        elif is_finite(min_value) and not is_finite(max_value):
+            score = 1.0 - np.exp(min_value - raw_score)
+
+        else:
+            raise ValueError('Unbounded score on both sides!')
```
fealho, line 99: Maybe I should use a sigmoid here instead of raising an error.

sdmetrics/base.py
```diff
@@ -58,3 +61,47 @@ def compute(real_data, synthetic_data):
...
+
+        Args:
+            raw_score (float):
+                The value of the metric from `compute`.
+
+        Returns:
+            float:
+                The normalized value of the metric
+        """
+        min_value = float(cls.min_value) if cls.min_value is not None else float('-inf')
```
fealho, line 77: Delete these 2 lines

sdmetrics/base.py
```diff
@@ -58,3 +61,47 @@ def compute(real_data, synthetic_data):
...
+        print(min_value)
+        print(max_value)
+        if max_value < raw_score or min_value > raw_score:
+            raise ValueError('`raw_score` must be between `min_value` and `max_value`.')
+
+        def is_finite(value):
+            return value not in (float('-inf'), float('inf'))
+
+        score = None
+        if is_finite(min_value) and is_finite(max_value):
```
fealho, line 89: Don't need the function.

sdmetrics/base.py
```diff
@@ -58,3 +61,47 @@ def compute(real_data, synthetic_data):
...
+            score = np.exp(raw_score - max_value)
+
+        elif is_finite(min_value) and not is_finite(max_value):
+            score = 1.0 - np.exp(min_value - raw_score)
+
+        else:
+            raise ValueError('Unbounded score on both sides!')
+
+        if score is None or score < 0 or score > 1:
+            raise ValueError(f'The score {score} should be a value between 0 and 1.')
```
fealho, line 102: Change error to AssertionError, and message to tell the user they did nothing wrong, we did something wrong.

pvk-developer (approved):

PR #68, opened 2021-07-16
Makes pomegranate optional

### LINKED ISSUE #63: Make pomegranate an optional dependency
The `pomegranate` dependency has very narrow sub-dependency ranges, which makes it easily incompatible with other libraries.
This provokes problems such as [this one when using SDV on Google Colab](https://github.com/sdv-dev/SDV/issues/417).

To avoid this, we can make `pomegranate` an optional dependency by using a lazy import inside the [bayesian_network.py](https://github.com/sdv-dev/SDMetrics/blob/master/sdmetrics/single_table/bayesian_network.py) module, which would raise an `ImportError` that invites the user to install `pomegranate` if not found.

fealho:
Resolves #63.

### DIFF
--- a/sdmetrics/single_table/bayesian_network.py
+++ b/sdmetrics/single_table/bayesian_network.py
@@ -4,7 +4,6 @@ import json
 import logging
 
 import numpy as np
-from pomegranate import BayesianNetwork
 
 from sdmetrics.goal import Goal
 from sdmetrics.single_table.base import SingleTableMetric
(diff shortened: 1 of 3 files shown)


katxiao (approved):
Should we add a Makefile target that installs pomegranate?

fealho:
@katxiao you mean create a make command that just installs pomegranate? If that's the case, I don't think it's necessary, since we might as well just pip install it directly. I'm merging this for now for the release, but we can open an issue to add it if you think appropriate.

katxiao:
> @katxiao you mean create a make command that just installs pomegranate? If that's the case, I don't think it's necessary, since we might as well just pip install it directly. I'm merging this for now for the release, but we can open an issue about this if you think it's appropriate.

Yeah that's what I mean. It might be convenient to have a `make install-extras` command or something. We can revisit it if it comes up again.

Issue #69, opened 2021-08-11
Upgrade dependency ranges


csala:
The latest versions of the libraries `pandas`, `sktime` and `pomegranate` are not supported by `sdmetrics`:

| Library | Upper bound (unsupported) | Latest release |
|-|-|-|
| `pandas` | 1.1.5 | 1.3.1 |
| `pomegranate` | 0.14.2 | 0.14.5 |
| `sktime` | 0.6 | 0.7.0 |

We should investigate why and update the code if necessary to support them.

katxiao:
https://github.com/sdv-dev/Copulas/issues/266 addresses `pandas`, and #81 addresses `sktime`. Since `pomegranate` is now marked as optional, I am going to close this issue. We can open another issue if we want to bump the optional pomegranate range after we stop supporting py36.

PR #81, opened 2021-10-29
Update sktime to support latest version

katxiao:
Update sktime to support the latest version as part of #69 

There was change in the import path for `TimeSeriesForestClassifier` starting at `v0.6.0`. We can update the dependency range to start at `v0.6.0` to support the newest versions. Also bump the min for `numpy` and `scikit-learn`, as required by `sktime v0.6.0`.


conda/meta.yaml
```diff
@@ -18,24 +18,24 @@ requirements:
     - pip
     - pytest-runner
     - python >=3.6,<3.9
-    - scikit-learn >=0.23,<1
+    - scikit-learn >=0.24,<2
```
amontanez24, line 21: Why is the new upper bound so high? I think for libraries lower than 1.0.0, we usually set the upper limit to one middle value higher. For example, here it would be `<0.3`.

katxiao: The newest version is `1.0.1` already - https://pypi.org/project/scikit-learn/#history

I don't see any reason to not take the latest version. We could do `<1.1` if that's preferred.

amontanez24: my understanding is that when the major version is `0`, sometimes the next update could be API breaking even if it only updates the minor version

katxiao: I see. But in this case we've already verified that it can use up to `1.0.1`.

csala: I agree with Katharine here: Now`scikit-learn` is already passed the 1.0 version, so we can expect them to not break the API until they advance to 2.x.

csala (approved):

amontanez24 (approved):

PR #82, opened 2021-10-29
Add pip check to workflows

### LINKED ISSUE #79: Add `pip check` to CI workflows
The current CI workflows and local test and lint commands do not catch dependency incompatibilities.

For example, installing the repository for development on the v0.6.0-dev branch results in these errors:
```
ERROR: numba 0.54.1 has requirement numpy<1.21,>=1.17, but you'll have numpy 1.21.3 which is incompatible.
ERROR: sphinx-rtd-theme 0.5.2 has requirement docutils<0.17, but you'll have docutils 0.18 which is incompatible.
ERROR: autopep8 1.6.0 has requirement pycodestyle>=2.8.0, but you'll have pycodestyle 2.7.0 which is incompatible.
```

A pip check command should be made part of the local and CI tests to make sure that our dependency tree is always clean.

pvk-developer:
* add pip check
* fix dependency errors
* Remove docs

Resolves #79


setup.py
```diff
@@ -15,7 +15,8 @@
     'numpy>=1.18.0,<2',
     'pandas>=1.1,<1.1.5',
     'scikit-learn>=0.23,<1',
-    'scipy>=1.4.1,<2',
+    'scipy>=1.4.1,<1.5',
```
katxiao, line 18: Why are we capping `scipy` at 1.5? I believe this will create conflicts with our other repos, since they support up to `2`.

pvk-developer: I reverted it, I changed it while I was trying to fix the dependencies  and forgot to roll it back.

tox.ini
```diff
@@ -23,12 +10,14 @@ deps =
...
+    integration: test
     minimum: test
 commands =
     lint: invoke lint
     readme: invoke readme
-    pytest: invoke pytest
+    unit: invoke unit
+    integration: invoke integration
     minimum: invoke minimum
     tutorials: invoke tutorials
```
csala, line 22: I do not think we have `tutorials` anymore, so maybe this and the other references to `tutorials` should be removed.

setup.py
```diff
@@ -16,6 +16,7 @@
     'pandas>=1.1,<1.1.5',
     'scikit-learn>=0.23,<1',
     'scipy>=1.4.1,<2',
+    'numba>=0.50,<0.54',
```
csala, line 19: Why is this needed?

csala (changes requested):
This looks good, but I think that the `tutorials` references can be removed altogether

setup.py
pvk-developer: `ERROR: numba 0.54.1 has requirement numpy<1.21,>=1.17, but you'll have numpy 1.21.3 which is incompatible.`
Which comes from `'sktime>=0.4,<0.6',` that do not cap `numba` https://github.com/alan-turing-institute/sktime/blob/main/setup.py which ends up pulling the latest release, that actually produces that error.

pvk-developer: Even with the new changes this still fails.

csala: I'm a bit confused about this @pvk-developer: So `numba==0.54.1` requires `numpy<1.21`, but `numba<0.54` works well with `numpy==1.21.3`?

pvk-developer: Yes, looks like numba `0.54` is the only version that does not accept that `numpy` version: [setup.py v 0.54](https://github.com/numba/numba/blob/release0.54/setup.py) vs [setup.py master](https://github.com/numba/numba/blob/master/setup.py) vs [setup.py v0.53](https://github.com/numba/numba/blob/release0.53/setup.py)

csala: I see. Let's keep it like this, then.

csala (approved):

PR #85, opened 2021-11-04
Fix lint after adding addon

fealho:
As part of #80, adds `flake8-pytest-style` and adapt the current code to pass this lint.


setup.cfg
```diff
@@ -34,7 +34,22 @@ universal = 1
...
+				D100, # Missing docstring in public module
+				D101, # Missing docstring in public class
+				D102, # Missing docstring in public method
+				D103, # Missing docstring in public function
+				D104, # Missing docstring in public package
+				D107, # Missing docstring in __init__
+				D205, # __init__ needs a description
+				D400, # __init__ needs a description
+				D401, # First line should be in imperative mood
+				VNE001, # single letter variable names like 'X' are not allowed
```
amontanez24, line 48: what's the reasoning for ignoring all of these?

PR #91, opened 2021-11-04
Add `flake8` addons that pass in the existing code

fealho:
As part of #80, add all `flake8` addons which do not require any code changes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,11 @@ universal = 1
 [flake8]
 max-line-length = 99
 exclude = docs, .tox, .git, __pycache__, .ipynb_checkpoints
+inline-quotes = single
+
+[pydocstyle]
+convention = google
+add-ignore = D107, D407, D417
 
 [isort]
 include_trailing_comment = True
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -34,6 +34,11 @@ universal = 1
 [flake8]
 max-line-length = 99
 exclude = docs, .tox, .git, __pycache__, .ipynb_checkpoints
+inline-quotes = single
+
+[pydocstyle]
+convention = google
+add-ignore = D107, D407, D417
```
csala, line 41: On a previous PR you used a different format that I liked a lot: multiple lines for the ignores, with a comment alongside each one mentioning what it was. I think that is very practical and makes it easier to maintain in the future.
Maybe we could do it here too?

fealho: Sure. I'm actually moving `pydocstyle` to its own PR though, because it doesn't pass in the existing code.

csala (approved):

pvk-developer (approved):
:shipit:

PR #94, opened 2021-11-04
Add `pep8-naming`

fealho:
As part of #80 added `pep8-naming` addon for `flake8` and adapted our code to follow this style.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ universal = 1
 [flake8]
 max-line-length = 99
 exclude = docs, .tox, .git, __pycache__, .ipynb_checkpoints
-ignore-names = X, Y, X_train, X_test, Y_col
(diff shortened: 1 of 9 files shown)


setup.cfg
```diff
@@ -34,6 +34,7 @@ universal = 1
 [flake8]
 max-line-length = 99
 exclude = docs, .tox, .git, __pycache__, .ipynb_checkpoints
+ignore-names = X, Y, X_train, X_test, Y_col
```
csala, line 37: Most of this names are very standard and acceptable, but I think this is not true for `Y_col`.
If it is not a huge change, I'd prefer to remove this from the list and adapt the corresponding code.

csala (approved):

pvk-developer (approved):
LGTM!

PR #95, opened 2021-11-04
Add `pandas-vet`

fealho:
As part of #80 , added `pandas-vet` and adapted the code.


sdmetrics/single_table/multi_single_column.py
```diff
@@ -70,8 +70,8 @@ def _compute(self, real_data, synthetic_data, metadata=None, **kwargs):
         scores = []
         for column_name, real_column in real_data.items():
             if column_name in fields:
-                real_column = real_column.values
-                synthetic_column = synthetic_data[column_name].values
+                real_column = real_column.array
```
csala, line 73: I see that some `values` were changed to `array` while the others changed to `to_numpy()`.
If I'm not mistaken, `array` works for `Series` only, so I suppose that is the reason, but what do you think about using `to_numpy()` everywhere, since it works for both `Series` and `DataFrame`?

katxiao (approved):

PR #96, opened 2021-11-04
Add `dlint`

fealho:
As part of #80 , added `dlint` and adapted the code.

### DIFF
--- a/sdmetrics/demos.py
+++ b/sdmetrics/demos.py
@@ -22,7 +22,7 @@ def load_demo(modality='multi_table'):
     """
     demo_path = pathlib.Path(__file__).parent / 'demos' / f'{modality}.pkl'
     with open(demo_path, 'rb') as demo_file:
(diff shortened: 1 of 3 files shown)


sdmetrics/demos.py
```diff
@@ -22,7 +22,7 @@ def load_demo(modality='multi_table'):
     """
     demo_path = pathlib.Path(__file__).parent / 'demos' / f'{modality}.pkl'
     with open(demo_path, 'rb') as demo_file:
-        return pickle.load(demo_file)
+        return pickle.load(demo_file)  # noqa: DUO103
```
csala, line 25: This is a very common use case within our code, loading the path that the user gave us, so I would rather move the ignore to the `setup.cfg` and make it standard

amontanez24 (approved):

csala (approved):

PR #97, opened 2021-11-04
Add `flake8-variables`

fealho:
As part of #80, add `flake8-variables` and change the code to conform with it.

**Note:** rule VNE001 is being ignored because it complains about single letter variables like X and Y. I could not find any way of whitelisting those variables specifically, so instead of putting tens of `noqa`'s, I'm simply ignoring the rule.


csala (approved):

pvk-developer (approved):

pvk-developer:
> I think that maybe we should change `y` and `X` to a better variable names if we want to keep usage of this plugin, because if we ignore the error then if a contributor comes with a variable like `a = 1` the lint will allow it to pass. What are your thoughts @fealho @csala ?

After a discussion with @fealho and @csala we can leave this as it is, ignore those for now and open an issue in the future for adding `pylint` which will end up covering some of this lints.

PR is approved 👍🏻

PR #98, opened 2021-11-04
Add `flake8-docstrings`

fealho:
As part of #80, add `flake8-docstrings` and update the code to conform with it.

**Note:** rule D107 (`__init__` docstrings) is being ignored, because in general we don't add docstrings to ``__init__``'s. This is also being ignored in RDT.

### DIFF
--- a/sdmetrics/single_table/privacy/loss.py
+++ b/sdmetrics/single_table/privacy/loss.py
@@ -1,3 +1,5 @@
+"""Utilities for the single_table.privacy modules."""
+
 import numpy as np
 from copulas.univariate.base import Univariate
 
@@ -36,7 +38,8 @@ class InverseCDFDistance(LossFunction):
     """
 
     def __init__(self, p=2):
-        """
+        """Set up the appropriate attributes.
+
         Args:
             p (float):
                 The p parameter in L_p metric. Must be positive.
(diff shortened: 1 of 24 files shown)


fealho:
@pvk-developer this addon is not in the list in issue #80, but it was added to RDT, which is why I added it here. Why is that? Did you forget to add it there, or is it just unnecessary?

sdmetrics/single_table/privacy/loss.py
```diff
@@ -36,7 +38,8 @@ class InverseCDFDistance(LossFunction):
     """
 
     def __init__(self, p=2):
-        """
+        """Set up the appropriate attributes.
```
csala, line 41: For `__init__` methods I think it's better to just not have a docstring at all and rely on the class docstring only, otherwise `Sphinx` has trouble rendering the right thing. So, what I would do is move the `Args` section to the class `docstring`, so Sphinx can find and render it, and remove this docstring altogether.
Notice that this will be accepted by the flake8 addon, so it won't complain about the `__init__` not having a docstring either.

sdmetrics/single_table/privacy/loss.py
```diff
@@ -33,14 +35,13 @@ class InverseCDFDistance(LossFunction):
     This loss function first applies the fitted cdfs to every single entry (i.e. turning
     the numerical values into their respective percentiles) and then measures the Lp distance
     to the pth power, between the predicted value and the real value.
+
```
katxiao, line 38: Is it our own convention to put the `args` in the class description instead of `__init__`?

katxiao (approved):
LGTM, just had a question.

csala (approved):

PR #99, opened 2021-11-05
Add `flake8-sfs`

fealho:
As part of #80, add `flake8-sfs` and update the code to match this addition.

**Note:** SFS3 (`String literal formatting using f-string`) is being ignored because we do so as well in RDT.


csala (approved):

amontanez24 (approved):
Looks good! Although could we leave a comment in the PR description about what `SFS3` is and the reason we ignore it?

csala (approved):

Issue #102, opened 2021-11-11
Add support for Python 3.9


pvk-developer:
SDMetrics package should now support Python 3.9

PR #105, opened 2021-11-23
Specify default categorical transformers

katxiao:
Update default data type transformers keys to the expected data types. Also specify `error_on_unknown=False` to preserve the original default [OneHotEncoderTransformer behavior](https://github.com/sdv-dev/RDT/blob/v0.5.3/rdt/hyper_transformer.py#L64).

Follow up from Issue #102


amontanez24 (approved):
LGTM!

PR #108, opened 2021-12-01
Capture metric errors in compute_metrics

### LINKED ISSUE #107: Report metric errors from `compute_metrics`
### Problem Description

Currently, `compute_metrics` does not capture what metrics are erroring out and what errors are being thrown. Instead, metrics that error out are being reported as `NaN`s and ultimately are ignored. As a result, end users have little information about what metrics are erroring out and why. Capturing this information could be useful for users of the library (such as `sdv.evaluate`) when debugging usage. 

### Expected behavior

We could add an `error` column to the final scores DataFrame.

katxiao:
Resolves #107 
Resolves [647](https://github.com/sdv-dev/SDV/issues/647)

### DIFF
--- a/sdmetrics/single_table/detection/base.py
+++ b/sdmetrics/single_table/detection/base.py
@@ -77,19 +77,15 @@ class DetectionMetric(SingleTableMetric):
         if np.isin(X, [np.inf, -np.inf]).any():
             X[np.isin(X, [np.inf, -np.inf])] = np.nan
 
-        try:
-            scores = []
-            kf = StratifiedKFold(n_splits=3, shuffle=True)
-            for train_index, test_index in kf.split(X, y):
-                y_pred = cls._fit_predict(X[train_index], y[train_index], X[test_index])
-                roc_auc = roc_auc_score(y[test_index], y_pred)
+        scores = []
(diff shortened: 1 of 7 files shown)


sdmetrics/single_table/detection/base.py
```diff
@@ -77,19 +77,15 @@ def compute(cls, real_data, synthetic_data, metadata=None):
...
-            scores = []
-            kf = StratifiedKFold(n_splits=3, shuffle=True)
-            for train_index, test_index in kf.split(X, y):
-                y_pred = cls._fit_predict(X[train_index], y[train_index], X[test_index])
-                roc_auc = roc_auc_score(y[test_index], y_pred)
-
-                scores.append(max(0.5, roc_auc) * 2 - 1)
-
-            return 1 - np.mean(scores)
-        except ValueError as err:
```
katxiao, line 90: It seems like `compute` should throw errors, and then we can catch them in `compute_metrics` and process the error as desired (e.g. continue onto other metrics)

csala (approved):
This looks good @katxiao !

fealho (approved):
Looks good 👍

PR #110, opened 2021-12-15
Fix multi table metric error handling

### LINKED ISSUE #109: Relational `KSTest` crashes with `IncomputableMetricError` if a table doesn't have numerical columns
### Environment Details

* SDMetrics version: 0.4.1
* Python version: 3.7

### Error Description

The relational `KSTest` is supposed to run the `KSTest` on all numerical columns in all tables and return the average score.

However, this test crashes if it encounters a table that has no numerical columns. I expect this test to succeed as long as there is at least 1 numerical column in any of the tables.

### Steps to reproduce

Use the relational demo dataset and pass it in with the metadata.

```python
from sdv.metrics.demos import load_multi_table_demo
from sdv.metrics.relational import KSTest

real_data, synthetic_data, metadata = load_multi_table_demo()
KSTest.compute(real_data, synthetic_data, metadata)
```

Output:
```
/usr/local/lib/python3.7/dist-packages/sdmetrics/single_table/base.py in _select_fields(cls, metadata, types)
     78 
     79         if len(fields) == 0:
---> 80             raise IncomputableMetricError(f'Cannot find fields of types {types}')
     81 
     82         return fields

IncomputableMetricError: Cannot find fields of types ('numerical',)
```

I believe this is happening because table `sessions` has no numerical columns. Interestingly, it _does_ work if I exclude the `metadata` object -- because then it starts assuming that the `id` field is a numerical column.

```python
KSTest.compute(real_data, synthetic_data)

0.8555555555555556
```

katxiao:
Multi table metrics should only error if all single table metrics have errored.

Resolves #109

### DIFF
--- a/sdmetrics/multi_table/multi_single_table.py
+++ b/sdmetrics/multi_table/multi_single_table.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 import numpy as np
 
 from sdmetrics import single_table
+from sdmetrics.errors import IncomputableMetricError
 from sdmetrics.multi_table.base import MultiTableMetric
 from sdmetrics.utils import nested_attrs_meta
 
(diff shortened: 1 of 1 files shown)


sdmetrics/multi_table/multi_single_table.py
```diff
@@ -67,12 +68,19 @@ def _compute(self, real_data, synthetic_data, metadata=None):
...
         for table_name, real_table in real_data.items():
             synthetic_table = synthetic_data[table_name]
             table_meta = metadata['tables'][table_name]
 
-            score = self.single_table_metric.compute(real_table, synthetic_table, table_meta)
-            values.append(score)
+            try:
+                score = self.single_table_metric.compute(real_table, synthetic_table, table_meta)
+                values.append(score)
+            except IncomputableMetricError as e:
```
csala, line 79: This should probably not be a single char variable. We can call it `error` or `ime`.

csala (approved):
The concept is fine. I just left a comment about a variable that we may want to rename.

csala (approved):

PR #117, opened 2022-01-28
Cap pomegranate to <14.7. Issue #116

### LINKED ISSUE #116: Cap pomegranate to <0.14.7
The latest `pomegranate` version, `0.14.7`, does not work well with `numpy<1.22` because of incompatibilities between the Python code and the compiled C backend.

When installed, SDMetrics currently ends up using `numpy==1.21.5` because of third party restrictions ([`numba` is not compatible with `numpy~=1.22` yet](https://github.com/numba/numba/issues/7754)), which means that SDMetrics crashes when `pomegranate` is used with the following error:

```
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

An example of this can be seen here: https://github.com/sdv-dev/SDMetrics/runs/4871113960?check_suite_focus=true

We can fix this by temporarily capping `pomegranate` on `<0.14.7` until `numba` extends support to the latest numpy version.

csala:
Resolve #116

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ install_requires = [
 
 pomegranate_requires = [
     "pomegranate>=0.13.4,<0.14.2;python_version<'3.7'",
-    "pomegranate>=0.14.1,<0.15;python_version>='3.7'",
+    "pomegranate>=0.14.1,<0.14.7;python_version>='3.7'",
 ]
 
 setup_requires = [


amontanez24 (approved):
:shipit:

pvk-developer (approved):
🚢 'it

PR #146, opened 2022-06-29
Compute score breakdown for `multi_single_column`

### LINKED ISSUE #130: Get `KSComplement` Score Breakdown
### Problem Description

As a user, I want to get a per-column breakdown of KS Scores when using this test on a tabular or multi-table dataset.

### Expected behavior

See #129 for background.

Create a new method `compute_breakdown` that has the same arguments as `compute` but returns a per-column breakdown instead of a summarized score.

Requirements:
- Return a dictionary of column names to scores. In the case of relational, it will be a nested dictionary.
- All column names should be present, even ones that are not the correct data type. Score should be `None` when the data type is not compatible
- If any other error occurs, the score should have the error object

### Examples

**Tabular**:
```python
from sdv.metrics.tabular import KSComplement 

KSComplement.compute_breakdown(real_data, synthetic_data, metadata)
{
  'age': 0.545656,
  'weight': 2343434,
  'gender': None, # the data type is categorical, which is not compatible
  'gpa': Error # this is a valid type but there was an error running it
}
```

**Relational**: Returned object is in the nested form, with the table names at the top
```python
from sdv.metrics.relational import KSComplement 

KSComplement.compute_breakdown(real_data, synthetic_data, metadata)
{
  'users': {
    'age': 0.545656,
    'weight': 2343434,
    'gender': None, # the data type is categorical, which is not compatible
    'gpa': Error # this is a valid type but there was an error running it
  }, 
 'transactions': {
    'transaction_id': None,
    'purchase_amt': 0.988191
 }
}
```

### Other Context

Eventually, we'll want to do the same thing for other metrics that are actually summarize of multiple scores

katxiao:
Resolves #130

### DIFF
--- a/sdmetrics/multi_table/multi_single_table.py
+++ b/sdmetrics/multi_table/multi_single_table.py
@@ -67,22 +67,28 @@ class MultiSingleTableMetric(MultiTableMetric, metaclass=nested_attrs_meta('sing
         elif not isinstance(metadata, dict):
             metadata = metadata.to_dict()
 
-        values = []
+        scores = {}
         errors = {}
         for table_name, real_table in real_data.items():
             synthetic_table = synthetic_data[table_name]
             table_meta = metadata['tables'][table_name]
 
             try:
-                score = self.single_table_metric.compute(real_table, synthetic_table, table_meta)
-                values.append(score)
-            except IncomputableMetricError as error:
+                if hasattr(self.single_table_metric, 'compute_breakdown'):
+                    score_breakdown = self.single_table_metric.compute_breakdown(
+                        real_table, synthetic_table, table_meta)
+                    scores[table_name] = score_breakdown
+                else:
+                    score = self.single_table_metric.compute(
+                        real_table, synthetic_table, table_meta)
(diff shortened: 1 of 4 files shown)


sdmetrics/multi_table/multi_single_table.py
```diff
@@ -67,22 +67,28 @@ def _compute(self, real_data, synthetic_data, metadata=None):
...
-                values.append(score)
-            except IncomputableMetricError as error:
+                if hasattr(self.single_table_metric, 'compute_breakdown'):
+                    score_breakdown = self.single_table_metric.compute_breakdown(
+                        real_table, synthetic_table, table_meta)
+                    scores[table_name] = score_breakdown
+                else:
+                    score = self.single_table_metric.compute(
+                        real_table, synthetic_table, table_meta)
+                    scores[table_name] = score
```
amontanez24, line 84: instead of this check could we catch the attribute error and do compute instead and then have the other exception block you have to catch everything else?

amontanez24 (approved):
Minor comment, but lgtm!

PR #147, opened 2022-06-29
Add statistic similarity metric

### LINKED ISSUE #137: Add `StatisticSimilarity` Metric
### Problem Description

Add a `StatisticSimilarity` metric, which takes in the actual statistic ('mean', 'std', 'median') as a parameter. For each column, compute the statistic on the real data (xr) and the synthetic data (xs). Subtract them and then normalize it using the min and max of the real data.

image

Base Unit: Per-column
Compatible data: Numerical, Datetime ( converted to Unix timestamp)
Missing values: Ignore

Edge cases: If the score <0, clip it to 0.

### Expected behavior

```
# single table case
>>> StatisticSimilarity.compute(real, synthetic, metadata,
      statistic='mean')
0.998

>>> StatisticSimilarity.compute(real, synthetic, metadata,
      statistic='median')
0.891

>>> StatisticSimilarity.compute(real, synthetic, metadata,
      statistic='std_dev')
0.870

# single table case
{
  'user_age': { 'real': 34, 'synthetic': 45, 'score': 0.89 },
  ...
}
```

katxiao:
Resolves #137

### DIFF
--- a/sdmetrics/single_column/statistical/__init__.py
+++ b/sdmetrics/single_column/statistical/__init__.py
@@ -2,8 +2,10 @@
 
 from sdmetrics.single_column.statistical.cstest import CSTest
 from sdmetrics.single_column.statistical.kscomplement import KSComplement
+from sdmetrics.single_column.statistical.statistic_similarity import StatisticSimilarity
 
 __all__ = [
     'CSTest',
-    'KSComplement'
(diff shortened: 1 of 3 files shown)


pvk-developer (approved):
LGTM ! :shipit:

PR #149, opened 2022-07-06
Add boundary adherence metric

### LINKED ISSUE #138: Add `BoundaryAdherence` metric
### Problem Description

Add a `BoundaryAdherence` metric. For each column, compute the number of rows in the synthetic data that are within the min and max bounds of the real data. Normalize this by the total # of rows so it's a score between 0 and 1.

Base Unit: Per-column
Compatible data: Numerical, Datetime ( converted to Unix timestamp)
Missing values: Ignore

### Expected behavior
```
# single table case
>>> BoundaryAdherence.compute_breakdown(real, synthetic, metadata)
{
  'user_age': { 'score': 0.89 },
  ...
}
```
please remove this section.-->

katxiao:
Resolves #138

### DIFF
--- a/sdmetrics/single_column/statistical/boundary_adherence.py
+++ b/sdmetrics/single_column/statistical/boundary_adherence.py
@@ -51,7 +51,7 @@ class BoundaryAdherence(SingleColumnMetric):
 
         valid = synthetic_data.between(real_data.min(), real_data.max())
 
-        return valid.sum() / len(real_data)
+        return valid.sum() / len(synthetic_data)
 
     @classmethod
     def normalize(cls, raw_score):


sdmetrics/single_column/statistical/boundary_adherence.py
```diff
@@ -0,0 +1,68 @@
...
+        real_data = pd.Series(real_data).dropna()
+        synthetic_data = pd.Series(synthetic_data).dropna()
+
+        if is_datetime(real_data):
+            real_data = pd.to_numeric(real_data)
+            synthetic_data = pd.to_numeric(synthetic_data)
+
+        valid = synthetic_data.between(real_data.min(), real_data.max())
+
+        return valid.sum() / len(real_data)
```
amontanez24, line 54: is there a reason the denominator isn't `len(synthetic_data)`. Are they the same length?

amontanez24 (approved):
I just have one question, but besides that looks good to go!

sdmetrics/single_column/statistical/boundary_adherence.py
katxiao: Oh yes, I'm assuming they are the same length, but I can switch it to `len(synthetic_data)` to be more safe & clear.

PR #154, opened 2022-07-11
Add missing value similarity metric

### LINKED ISSUE #139: Add `MissingValueSimilarity` metric
### Problem Description

Add a `MissingValueSimilarity` metric. For each column, compute the % of missing values in the real data (mr) as a number between [0,1]; do the same for the synthetic data (sr). The score is the absolute value of the difference between the real and synthetic data. Normalize it.
image

The breakdown should include the % missing values in the real vs. synthetic data, as well as the overall score.

### Expected behavior

```
# single table case
>>> MissingValueSimilarity.compute_breakdown(real, synthetic, metadata)
{
  'user_age': { 'real': 0.2, 'synthetic': 0.6, 'score': 0.4 },
  ...
}
```

katxiao:
Resolves #139

### DIFF
--- a/sdmetrics/single_column/statistical/missing_value_similarity.py
+++ b/sdmetrics/single_column/statistical/missing_value_similarity.py
@@ -25,22 +25,6 @@ class MissingValueSimilarity(SingleColumnMetric):
     min_value = 0.0
     max_value = 1.0
 
-    def compute(self, real_data, synthetic_data):
-        """Compare the missing value similarity of two continuous columns.
-
-        Args:
-            real_data (Union[numpy.ndarray, pandas.Series]):
-                The values from the real dataset.
-            synthetic_data (Union[numpy.ndarray, pandas.Series]):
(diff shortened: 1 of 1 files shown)


sdmetrics/single_column/statistical/missing_value_similarity.py
```diff
@@ -0,0 +1,81 @@
...
+        synthetic_data = pd.Series(synthetic_data)
+
+        real_data_value = real_data.isna().sum() / len(real_data)
+        synthetic_data_value = synthetic_data.isna().sum() / len(synthetic_data)
+
+        return {
+            'score': 1 - abs(real_data_value - synthetic_data_value),
+            'real': real_data_value,
+            'synthetic': synthetic_data_value,
+        }
```
pvk-developer, line 67: minor: Could we move this code before the `compute` since the `compute` is calling this method.

pvk-developer (approved):

PR #155, opened 2022-07-11
Add category coverage metric

### LINKED ISSUE #140: Add `CategoryCoverage` metric
### Problem Description

Add a `CategoryCoverage` metric. For each column: Look at the # of unique categories present in the real data (cr). Then, calculate the number of those that are present in the synthetic data (cs); to be present means that there is at least 1 row with that value.

image

Base Unit: Per-column
Compatible data: Categorical (not pii), Boolean (not pii)
Missing values: Treat missing values as if they are 1 single, additional category

The breakdown should include some intermediate outputs:
Number of distinct categories in real data
Number of those categories in the synthetic data. Note if the synthetic data happens to contain new, unseen categories, do not count those.
Score

### Expected behavior

```
# single table case
>>> CategoryCoverage.compute_breakdown(real, synthetic, metadata)
{
  'user_age': { 'real': 30, 'synthetic': 29, 'score': 0.96667 },
  ...
}
```

katxiao:
Resolves #140

### DIFF
--- a/sdmetrics/single_column/statistical/category_coverage.py
+++ b/sdmetrics/single_column/statistical/category_coverage.py
@@ -54,8 +54,8 @@ class CategoryCoverage(SingleColumnMetric):
             dict:
                 A mapping of the category coverage results.
         """
-        real_data = pd.Series(real_data)
-        synthetic_data = pd.Series(synthetic_data)
+        real_data = pd.Series(real_data).dropna()
+        synthetic_data = pd.Series(synthetic_data).dropna()
 
         real_data_value = real_data.nunique()
         synthetic_data_value = synthetic_data.nunique()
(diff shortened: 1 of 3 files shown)


sdmetrics/single_column/statistical/category_coverage.py
```diff
@@ -0,0 +1,81 @@
...
+                A mapping of the category coverage results.
+        """
+        real_data = pd.Series(real_data).dropna()
+        synthetic_data = pd.Series(synthetic_data).dropna()
+
+        real_data_value = real_data.nunique()
+        synthetic_data_value = synthetic_data.nunique()
+
+        return {
+            'score': synthetic_data_value / real_data_value,
```
amontanez24, line 64: should we do something to protect against empty arrays and potential divide by zeros?

amontanez24 (approved):
One question about potential divide by 0 errors but besides that LGTM!

sdmetrics/single_column/statistical/category_coverage.py
katxiao: I think we would prefer to throw the error in this case and surface it in the summary. In the future, we can introduce a more comprehensive validation that can return very user-friendly errors.

PR #157, opened 2022-07-13
Add total variation complement

### LINKED ISSUE #142: Add `TVComplement` metric
### Problem Description

Add a `TVComplement` metric. For each discrete column: Calculate the frequencies of each category in the real data (r0, r1, …), expressed as a proportion in [0, 1]; also calculate frequencies in synthetic data (s0, s1, …). Calculate the [Total Variation Distance](https://en.wikipedia.org/wiki/Total_variation_distance_of_probability_measures) between the two distributions and then normalize it.

image

The breakdown only contains the score.

### Expected behavior

```
# single table case
>>> TVComplement.compute_breakdown(real, synthetic, metadata)
{
  'user_age': { 'score': 0.96667 },
  ...
}
```

katxiao:
Resolves #142


fealho (approved):
💯

PR #162, opened 2022-07-14
Add contingency similarity metric

### LINKED ISSUE #159: Add `ContingencySimilarity` Metric
### Problem Description

Base Unit: A pair of columns from the same table
Overall Summary: Average score across all pairs

Compatible data: Categorical, Boolean
Missing values: Treat as a single, new category

For each pair of columns (A, B):
1. Compute the contingency table in the synthetic table S and the real table R, where contingency table is the frequencies of every pair of categories in the real data (ai, bj) as a value between [0, 1].
2. Measure the difference between the contingency tables using the Total Variation Distance metric
3. Return 1-TVD so that a higher score = higher similarity

The breakdown is keyed by the column pairs, represented as a tuple in alphabetical order. The value is the real correlation, synthetic correlation and the final score.

### Expected behavior
```
# single table case
>>> ContingencySimilarity.compute_breakdown(real, synthetic, metadata)
{
  ('user_age', 'user_purchase_amt'): { 'real': -0.451, 'synthetic': -0.003, 'score': 0.776 },
  ...
}

# for multi table, it's nested by table name
>>> TVComplement.compute_breakdown(real, synthetic, metadata)
{
  'users': {
    ('user_age', 'user_purchase_amt'): { 'score': 0.776 },
  }
   ...
}
```

katxiao:
Resolves #159

### DIFF
--- a/sdmetrics/column_pairs/statistical/__init__.py
+++ b/sdmetrics/column_pairs/statistical/__init__.py
@@ -1,9 +1,11 @@
 """Statistical Metrics to compare column pairs."""
 
+from sdmetrics.column_pairs.statistical.contingency_similarity import ContingencySimilarity
 from sdmetrics.column_pairs.statistical.kl_divergence import (
     ContinuousKLDivergence, DiscreteKLDivergence)
 
 __all__ = [
+    'ContingencySimilarity',
     'ContinuousKLDivergence',
     'DiscreteKLDivergence',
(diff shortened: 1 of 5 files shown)


pvk-developer (changes requested):
The file `test_contingency_similarity` should be in the `test` folders, other than that I think it looks good.

PR #165, opened 2022-07-15
Add cardinality statistic similarity metric

### LINKED ISSUE #145: Add `CardinalityStatisticSimilarity` metric
### Problem Description

Add a `CardinalityStatisticSimilarity` method for multi-table datasets, which takes the desired statistic calculation (average, standard deviation, median) as a parameter.

1. In the real data, note down the # of children each parent row has to get r1, r2, r3, …
2. Do the same for the synthetic data to get s1, s2, s3, …
3. Now that you have 2 distributions of discrete data, calculate the statistic on the real data to get xr and the synthetic data to get xs
4. Normalize the statistic by the min and max value of the real data.5. 

image

Edge cases: If the score <0, clip it to 0.

The breakdown will have a tuple of table names as keys (parent, child). The value will be the real statistic, synthetic statistic and the final value.

### Expected behavior

```
# single table case
>>> CardinalitySimilarity.compute(real, synthetic, metadata,
      statistic='average')
0.998

>>> CardinalitySimilarity.compute(real, synthetic, metadata,
      statistic='median')
0.891

>>> CardinalitySimilarity.compute(real, synthetic, metadata,
      statistic='std_dev')
0.870

# this metric is only available for multi table
{
  ('users', 'sessions'): {'real': 2.456, 'synthetic': 1.234, 'score': 0.8642 },
  ('sessions', 'transactions'): {'real': 2.0, 'synthetic': 2.5, 'score': 0.9 },
}
```

katxiao:
Resolves #145

### DIFF
--- a/sdmetrics/multi_table/statistical/cardinality_statistic_similarity.py
+++ b/sdmetrics/multi_table/statistical/cardinality_statistic_similarity.py
@@ -1,0 +1,182 @@
+"""The CardinalityStatisticSimilarity metric."""
+
+import numpy as np
+import pandas as pd
+
+from sdmetrics.goal import Goal
+from sdmetrics.multi_table.base import MultiTableMetric
+
+
+class CardinalityStatisticSimilarity(MultiTableMetric):
+    """CardinalityStatisticSimilarity metric computes the similarity of the cardinality.
+
+    Attributes:
+        name (str):
+            Name to use when reports about this metric are printed.
+        goal (sdmetrics.goal.Goal):
+            The goal of this metric.
+        min_value (float):
+            Minimum value that this metric can take.
+        max_value (float):
+            Maximum value that this metric can take.
+    """
+
+    name = 'CardinalityStatisticSimilarity'
+    goal = Goal.MAXIMIZE
+    min_value = 0.0
+    max_value = 1.0
+
+    @classmethod
+    def _get_cardinality_distribution(cls, parent_column, child_column):
+        """Compute the cardinality distribution of the (parent, child) pairing.
+
+        Args:
+            parent_column (pandas.Series):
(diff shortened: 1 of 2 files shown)


sdmetrics/utils.py
```diff
@@ -72,6 +72,26 @@ def get_frequencies(real, synthetic):
     return f_obs, f_exp
 
 
+def get_cardinality_distribution(parent_column, child_column):
```
amontanez24, line 75: any reason to put this in utils? Are we going to reuse it?

sdmetrics/multi_table/statistical/cardinality_statistic_similarity.py
```diff
@@ -0,0 +1,165 @@
...
+                real_data fields and dtypes.
+            statistic (str):
+                The desired statistic to compute. Must be either 'mean', 'median', or 'std'.
+
+        Returns:
+            float:
+                The average of all (parent, child) cardinality statistic similarity scores.
+        """
+        score_breakdowns = cls.compute_breakdown(real_data, synthetic_data, metadata, statistic)
+        if 'score' in score_breakdowns:
```
amontanez24, line 146: What is the situation where score ends up on the outside dict?

amontanez24 (commented):
Looking good! I just have some questions

sdmetrics/multi_table/statistical/cardinality_statistic_similarity.py
katxiao: It's if there are no relationships in the metadata (the metric is not applicable). In that case we set the metric score to `np.nan`.

amontanez24 (approved):
LGTM!

PR #166, opened 2022-07-15
Add cardinality shape similarity metric

### LINKED ISSUE #160: Add `CardinalityShapeSimilarity` metric
### Problem Description

Base Unit: A parent-child table relationship. This metric is only available for multi table datasets.
Overall Summary: Average score across all parent-child tables

1. In the real data, note down the # of children each parent row has to get r1, r2, r3, …
2. Do the same for the synthetic data to get s1, s2, s3, …
3. Now that you have 2 distributions of discrete data, calculate & return the KSComplement of the two distributions.4. 

Breakdown will have a tuple of table names as keys (parent, child). The value will be the real statistic, synthetic statistic and the final value.

### Expected behavior
```
# this metric is only available for multi table
{
  ('users', 'sessions'): { 'score': 0.8642 },
  ('sessions', 'transactions'): { 'score': 0.9 },
}
```

katxiao:
Resolves #160

### DIFF
--- a/sdmetrics/multi_table/statistical/cardinality_shape_similarity.py
+++ b/sdmetrics/multi_table/statistical/cardinality_shape_similarity.py
@@ -1,0 +1,124 @@
+"""The CardinalityShapeSimilarity metric."""
+
+import numpy as np
+from scipy.stats import ks_2samp
+
+from sdmetrics.goal import Goal
+from sdmetrics.multi_table.base import MultiTableMetric
+from sdmetrics.utils import get_cardinality_distribution
+
+
+class CardinalityShapeSimilarity(MultiTableMetric):
+    """CardinalityShapeSimilarity metric computes the similarity of the cardinality.
+
+    Attributes:
+        name (str):
+            Name to use when reports about this metric are printed.
+        goal (sdmetrics.goal.Goal):
+            The goal of this metric.
+        min_value (float):
+            Minimum value that this metric can take.
+        max_value (float):
+            Maximum value that this metric can take.
+    """
+
+    name = 'CardinalityShapeSimilarity'
+    goal = Goal.MAXIMIZE
+    min_value = 0.0
+    max_value = 1.0
+
+    @classmethod
+    def compute_breakdown(cls, real_data, synthetic_data, metadata=None):
+        """Compute the breakdown of cardinality shape similarity in the given tables.
+
+        Compute the cardinality distributions for the real and synthetic data for each
+        (parent, child) relationship specified in the metadata. Then compute the KSComplement
+        over the two cardinality distributions, and compare the final scores
+        to obtain the cardinality shape similarity score.
+
+        Args:
+            real_data (dict[str, pandas.DataFrame]):
+                The tables from the real dataset, passed as a dictionary of
+                table names and pandas.DataFrames.
+            synthetic_data (dict[str, pandas.DataFrame]):
+                The tables from the synthetic dataset, passed as a dictionary of
+                table names and pandas.DataFrames.
+            metadata (dict):
+                Multi-table metadata dict. If not passed, it is build based on the
+                real_data fields and dtypes.
+
+        Returns:
+            dict:
+                A dict mapping (parent, child) values to the score breakdown for the
+                cardinality distributions of that foreign key.
+        """
+        if set(real_data.keys()) != set(synthetic_data.keys()):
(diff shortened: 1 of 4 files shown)


sdmetrics/multi_table/statistical/cardinality_shape_similarity.py
```diff
@@ -0,0 +1,127 @@
...
+
+        Returns:
+            dict:
+                A dict mapping (parent, child) values to the score breakdown for the
+                cardinality distributions of that foreign key.
+        """
+        if set(real_data.keys()) != set(synthetic_data.keys()):
+            raise ValueError('`real_data` and `synthetic_data` must have the same tables.')
+        if metadata is None:
+            raise ValueError('`metadata` cannot be ``None``.')
```
pvk-developer, line 58: If we require `metadata` then change the `metadata=None` to just `metadata`, that way it must always be passed. Then we don't need to do this check.

sdmetrics/multi_table/statistical/cardinality_shape_similarity.py
```diff
@@ -0,0 +1,127 @@
...
+    goal = Goal.MAXIMIZE
+    min_value = 0.0
+    max_value = 1.0
+
+    @classmethod
+    def compute_breakdown(cls, real_data, synthetic_data, metadata=None):
+        """Compute the breakdown of cardinality shape similarity in the given tables.
+
+        Compute the cardinality distributions for the real and synthetic data for each
+        (parent, child) relationship specified in the metadata. Then compute the KSComplement
```
pvk-developer, line 35: Use ``` `KSComplement` ``` for the docs. (I'm not sure if single or double ` )

sdmetrics/multi_table/statistical/cardinality_shape_similarity.py
```diff
@@ -0,0 +1,127 @@
...
+        Args:
+            real_data (dict[str, pandas.DataFrame]):
+                The tables from the real dataset, passed as a dictionary of
+                table names and pandas.DataFrames.
+            synthetic_data (dict[str, pandas.DataFrame]):
+                The tables from the synthetic dataset, passed as a dictionary of
+                table names and pandas.DataFrames.
+            metadata (dict):
+                Multi-table metadata dict. If not passed, it is build based on the
+                real_data fields and dtypes.
```
pvk-developer, line 48: This is a bit confusing since below we raise a `ValueError` if it's `None` .

sdmetrics/multi_table/statistical/cardinality_shape_similarity.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return score_breakdowns
+
+    @classmethod
+    def compute(cls, real_data, synthetic_data, metadata=None):
+        """Compute the average of cardinality shape similarity in the given tables.
+
+        Compute the average shape similarity in cardinality distributions for
+        (parent, child) relationship in the given tables, as specified in the metadata.
+        The shape similarity is computed based on the KSComplement.
```
pvk-developer, line 90: Use ``` `KSComplement` ``` for the docs.

sdmetrics/multi_table/statistical/cardinality_shape_similarity.py
```diff
@@ -0,0 +1,127 @@
...
+        Args:
+            real_data (dict[str, pandas.DataFrame]):
+                The tables from the real dataset, passed as a dictionary of
+                table names and pandas.DataFrames.
+            synthetic_data (dict[str, pandas.DataFrame]):
+                The tables from the synthetic dataset, passed as a dictionary of
+                table names and pandas.DataFrames.
+            metadata (dict):
+                Multi-table metadata dict. If not passed, it is build based on the
+                real_data fields and dtypes.
```
pvk-developer, line 101: Can this default to `None` ? I think we don't build the metadata anywhere. If it's `None` the `compute_breakdown` raises an error.

pvk-developer (changes requested):

sdmetrics/multi_table/statistical/cardinality_shape_similarity.py
katxiao: Hm yeah we could do that, but I was wondering if it's confusing since the parent method we are overriding has `metadata` as optional. I guess it's fine since we are effectively enforcing this rule anyway.

pvk-developer (approved):
LGTM!

PR #169, opened 2022-07-21
Add wrapper classes for new single and multi table metrics

katxiao:
Add new metrics to the single- and multi-table levels.


amontanez24 (approved):
LGTM!

PR #205, opened 2022-09-02
Add multi table quality report visualizations

### LINKED ISSUE #192: Add multi table QualityReport visualization methods
### Problem Description

Add multi table QualityReport visualization methods. These methods are available in the report object, even if it's saved and loaded at a later time.

Add the following methods:
- show_score()
- show_details(property_name)

katxiao:
Resolves #192


sdmetrics/reports/multi_table/plot_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    plot_data = _get_table_relationships_data(score_breakdowns)
+    average_score = round(plot_data['Quality Score'].mean(), 2)
+
+    fig = px.bar(
+        plot_data,
+        x='Child → Parent Relationship',
+        y='Quality Score',
+        title=f'Column Shapes Similarity (Average={average_score})',
+        color='Metric',
+        color_discrete_sequence=['#000036'],
```
amontanez24, line 55: minor: can we make the color codes like this and the one on line 68 constants?

sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -109,6 +111,43 @@ def get_properties(self):
...
+                The table to show scores for. Must be provided for 'Column Shapes'
+                and 'Column Pair Trends'
+        """
+        if property_name in ['Column Shapes', 'Column Pair Trends'] and table_name is None:
+            raise ValueError('Table name must be provided when viewing details for '
+                             f'property {property_name}.')
+
+        if property_name == 'Column Shapes':
+            score_breakdowns = {
+                metric.__name__: self._metric_results[metric.__name__].get(table_name, {})
```
amontanez24, line 130: minor: maybe we can pull out `metric.__name__` and `self._metric_results[metric.__name__]` into variables outside this if else block since it seems like all the statements use them

amontanez24 (approved):
a couple of minor comments but besides that :shipit:

sdmetrics/reports/multi_table/quality_report.py
katxiao: Oh, that's a little hard to do since `metric` is only defined in the inline for loop.

PR #214, opened 2022-09-07
Update column pair metric calculation

katxiao:
Discretize numerical and datetime columns so that we can calculate correlations between numerical & datetime columns with categorical columns.


fealho (approved):
👍

PR #219, opened 2022-09-09
Bug fixes for QualityReport

katxiao:
Formatting updates:
- Update `get_raw_result` output formatting
- Filter out null scores from `get_raw_result` output
- In `plot_column` util, make NaNs a category
- Make plot titles consistent & update subplot titles
- Format overall score as a percentage, and fix formatting for NaN scores
- Separate column pair tuple into two columns in `get_details`
- Update heatmap colors

Functionality fixes:
- Fix duplicate key bug for Column Pair Trends
- Fix missing value inconsistencies for real and synthetic correlation plots
- Handle edge case in single-table QualityReport where one metric has no results.
- Handle edge case in multi-table QualityReport where there is only one table (no parent-child relationship results).
- Multi table QualityReport's `show_details` and `get_details` should filter parent-child relationships by table name

Functionality updates:
- sort column pair tuples in alphabetical order within each tuple

### DIFF
--- a/sdmetrics/reports/multi_table/quality_report.py
+++ b/sdmetrics/reports/multi_table/quality_report.py
@@ -268,13 +268,20 @@ class QualityReport():
         metrics = list(itertools.chain.from_iterable(self.METRICS.values()))
         for metric in metrics:
             if metric.__name__ == metric_name:
+                filtered_results = {}
+                for table_name, table_results in self._metric_results[metric_name].items():
+                    filtered_results[table_name] = {
+                        key: result for key, result in table_results.items()
+                        if not np.isnan(result['score'])
+                    }
+
                 return [
                     {
                         'metric': {
                             'method': f'{metric.__module__}.{metric.__name__}',
                             'parameters': {},
                         },
-                        'results': self._metric_results[metric_name],
+                        'results': filtered_results,
                     },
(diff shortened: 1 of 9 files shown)


sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -90,13 +91,16 @@ def generate(self, real_data, synthetic_data, metadata):
...
-                            in self._metric_results['CardinalityShapeSimilarity'].items()
-                        ]
-                    )
-                    prop_scores.append(score)
+                    if 'score' in self._metric_results[metric.__name__]:
+                        prop_scores.append(self._metric_results[metric.__name__]['score'])
+                    else:
+                        score = np.nanmean(
+                            [
+                                table_breakdown['score'] for _, table_breakdown
```
pvk-developer, line 99: Should we use `table_breakdown.get('score', np.nan)` to prevent failed metrics ?

pvk-developer (approved):

sdmetrics/reports/multi_table/quality_report.py
katxiao: I'll address this in a follow-up PR since we want to do some additional error handling for this scenario.

PR #222, opened 2022-09-12
Parse date as datetime

katxiao:
Parse date data as pandas datetime in all single-table metrics.


fealho (commented):
You mentioned that this fixes a bug found in the bug bash, correct? Maybe add an integration test to check that that case was indeed covered.

Besides that, looks good!

katxiao:
> You mentioned that this fixes a bug found in the bug bash, correct? Maybe add an integration test to check that that case was indeed covered.
> 
> Besides that, looks good!

@fealho I added integration tests to check for date inputs.

fealho (approved):
💯

PR #225, opened 2022-09-14
Add range coverage metric

### LINKED ISSUE #208: Add RangeCoverage Metric
### Problem Description

Add a RangeCoverage metric that measures whether a synthetic column covers the full range of values that are present in a real column.

Data types: numerical, datetime
Ignore any missing values in numerical or datetime columns.

If r and s represent the real and synthetic columns, then this metric computes how close the min and max values of s come to the true min and max values in r according to the formula below
image

### Expected behavior

```
from sdmetrics.single_column import RangeCoverage

RangeCoverage.compute(
    real_data=real_column,
    synthetic_data=synthetic_column
)
```

katxiao:
Resolves #208

### DIFF
--- a/sdmetrics/multi_table/__init__.py
+++ b/sdmetrics/multi_table/__init__.py
@@ -38,4 +38,5 @@ __all__ = [
     'MissingValueSimilarity',
     'StatisticSimilarity',
     'TVComplement',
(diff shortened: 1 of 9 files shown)


pvk-developer (approved):
LGTM!

PR #226, opened 2022-09-15
Add new row synthesis single table metric

### LINKED ISSUE #207: Add NewRowSynthesis Metric
### Problem Description

Add a NewRowSynthesis metric that measures whether each row in the synthetic data is unique, or whether it exactly matches a row in the real data. This metric looks for matches between the real and synthetic data for the compatible columns. This metric also looks for matches in missing values.

Data types: Categorical, boolean, numerical, datetime

This metric looks for matching rows between the real and synthetic dataset. In order to be considered a match, all the individual values in the real row must match the synthetic row. The exact matching criteria is based on the type of data.
Categorical/Boolean Data: The value in the real data must be exactly the same as the value in the synthetic data. 
Numerical/Datetime Data: This metric scales every value in the real and synthetic data (x). This is shown in the formula below, where r represents all the values in the real data.
image
Then, we consider a match when the synthetic value is within a % of the real value. The % is is a parameter, which is set to 0.01 (1%) by default.

Missing Values: To be considered a match, both real and synthetic values must be missing.

Finally, we compute the proportion of rows in the synthetic data that match to a row in the real data. The score is the complement, ensuring that 1 is a good score (every row is unique) while 0 is the worst (every row has a match).

image

### Expected behavior

```
from sdmetrics.single_table import NewRowSynthesis

NewRowSynthesis.compute(
    real_data=real_columns,
    synthetic_data=synthetic_columns,
    numerical_match_tolerance=0.01,
    synthetic_sample_size=10_000
)
```

katxiao:
Resolves #207

### DIFF
--- a/sdmetrics/reports/single_table/plot_utils.py
+++ b/sdmetrics/reports/single_table/plot_utils.py
@@ -85,6 +85,8 @@ def _get_similarity_correlation_matrix(score_breakdowns, columns):
     Args:
         score_breakdowns (dict):
             Mapping of metric to the score breakdown result.
+        columns (list[string] or set[string]):
+            A list or set of column names.
 
     Returns:
         pandas.DataFrame
--- a/sdmetrics/reports/single_table/plot_utils.py
+++ b/sdmetrics/reports/single_table/plot_utils.py
@@ -89,6 +89,9 @@ def _get_similarity_correlation_matrix(score_breakdowns, columns):
     Returns:
         pandas.DataFrame
     """
+    if isinstance(columns, set):
+        columns = list(columns)
+
     similarity_correlation = pd.DataFrame(
         index=columns,
         columns=columns,
--- a/sdmetrics/single_table/new_row_synthesis.py
+++ b/sdmetrics/single_table/new_row_synthesis.py
@@ -1,6 +1,7 @@
 """New Row Synthesis metric for single table."""
 import warnings
 
+import numpy as np
 import pandas as pd
 
 from sdmetrics.goal import Goal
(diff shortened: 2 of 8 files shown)


sdmetrics/reports/single_table/plot_utils.py
```diff
@@ -89,6 +89,9 @@ def _get_similarity_correlation_matrix(score_breakdowns, columns):
     Returns:
         pandas.DataFrame
     """
+    if isinstance(columns, set):
+        columns = list(columns)
```
pvk-developer, line 93: Minor, but not sure what `columns` can be:
You can always cast to `list`  and this will be one dimensional:
```python
In [1]: columns_list = ['a', 'b', 'c']
In [2]: columns_set = {'a', 'b', 'c'}
In [3]: columns_tuple = ('a', 'b', 'c')

In [4]: list(columns_list)
Out[4]: ['a', 'b', 'c']

In [5]: list(columns_set)
Out[5]: ['b', 'c', 'a']

In [6]: list(columns_tuple)
Out[6]: ['a', 'b', 'c']
```
If `columns` can be a `str` then don't do this as it will split it in letters

sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Args:
+            real_data (Union[numpy.ndarray, pandas.DataFrame]):
+                The values from the real dataset.
+            synthetic_data (Union[numpy.ndarray, pandas.DataFrame]):
+                The values from the synthetic dataset.
+            metadata (dict):
+                Table metadata dict.
+            numerical_match_tolerance (float):
+                A float >0.0 representing how close two numerical values have to be
```
pvk-developer, line 49: could we use `bigger than 0` instead of `>0.0` and `Defaults to `0.01` at the end?

sdmetrics/reports/single_table/plot_utils.py
katxiao: It should always be a list or a set.

pvk-developer (approved):
This looks good now! :shipit:

PR #230, opened 2022-09-16
Fix column pairs visualization bug

katxiao:
Skip nan scores in column pairs visualization

### DIFF
--- a/sdmetrics/reports/single_table/plot_utils.py
+++ b/sdmetrics/reports/single_table/plot_utils.py
@@ -117,7 +117,11 @@ def _get_numerical_correlation_matrices(score_breakdowns):
             The real and synthetic numerical correlation matrices.
     """
     columns = []
-    for cols, _ in score_breakdowns['CorrelationSimilarity'].items():
+    valid_score_breakdowns = {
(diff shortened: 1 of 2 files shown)


sdmetrics/reports/single_table/plot_utils.py
```diff
@@ -117,7 +117,11 @@ def _get_numerical_correlation_matrices(score_breakdowns):
             The real and synthetic numerical correlation matrices.
     """
     columns = []
-    for cols, _ in score_breakdowns['CorrelationSimilarity'].items():
+    valid_score_breakdowns = {
+        k: breakdown for k, breakdown in score_breakdowns['CorrelationSimilarity'].items()
```
pvk-developer, line 121: Could we use `key` instead of `k`?
I think you can write this in a break line such as this:
```python
valid_score_breakdowns = {
    key: breakdown
    for key, breakdown in score_breakdowns['CorrelationSimilarity'].items()
    if not np.isnan(breakdown.get('score', np.nan))  # not sure if  get is needed or we always have `score` 
}
```

pvk-developer (approved):
Just minor nit but LGTM! 👍🏻

PR #233, opened 2022-09-26
Remove torch dependency

katxiao:
Remove sdmetric's torch dependency

Import it in the metric that uses it, and throw an error if not installed.


fealho (commented):
Looks good. I'll approve as soon as it passes all the tests

fealho (approved):

PR #238, opened 2022-09-30
Add single table diagnostic report visualizations

### LINKED ISSUE #211: Visualization methods for the single-table DiagnosticReport
### Problem Description

Add visualization methods for each of the properties in the single-table DiagnosticReport.
- show_details(property_name)

katxiao:
Resolves #211

### DIFF
--- a/sdmetrics/reports/single_table/diagnostic_report.py
+++ b/sdmetrics/reports/single_table/diagnostic_report.py
@@ -12,6 +12,8 @@ import pkg_resources
 import tqdm
 
 from sdmetrics.errors import IncomputableMetricError
+from sdmetrics.reports.single_table.plot_utils import (
+    get_column_boundaries_plot, get_column_coverage_plot, get_synthesis_plot)
 from sdmetrics.reports.utils import aggregate_metric_results
 from sdmetrics.single_table import (
     BoundaryAdherence, CategoryCoverage, NewRowSynthesis, RangeCoverage)
@@ -78,6 +80,7 @@ class DiagnosticReport():
     def __init__(self):
         self._metric_results = {}
         self._metric_averages = {}
+        self._property_scores = {}
         self._results = {}
 
     def _print_results_for_level(self, out, level):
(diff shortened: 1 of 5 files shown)


sdmetrics/reports/single_table/diagnostic_report.py
```diff
@@ -178,6 +181,22 @@ def get_visualization(self, property_name):
...
+        if property_name == 'Coverage':
+            fig = get_column_coverage_plot(score_breakdowns, self._property_scores[property_name])
+
+        elif property_name == 'Boundaries':
+            fig = get_column_boundaries_plot(
+                score_breakdowns, self._property_scores[property_name])
+
+        elif property_name == 'Synthesis':
+            fig = get_synthesis_plot(score_breakdowns.get('NewRowSynthesis', {}))
+
```
pvk-developer, line 198: I think we should use `property_name.lower() == 'coverage'` and ensure that even if the user doesn't write 'Coverage' with capital `c` this is being properly shown (same for the other cases).

sdmetrics/reports/single_table/diagnostic_report.py
```diff
@@ -178,6 +181,22 @@ def get_visualization(self, property_name):
...
+        return fig
```
pvk-developer, line 199: I think it would be nice to have a warning if `fig` is `None`. (I think we should define `fig = None` at the top, then if `fig` is `None` let the user know the allowed properties. If a `property` that is not one of the three above, we produce an `UnboundLocalError`.

pvk-developer (changes requested):
I left some suggestions, let me know if you agree with them! Other than that this is looking very good! 👍🏻

sdmetrics/reports/single_table/diagnostic_report.py
katxiao: Hm, I'll run this by Neha and if we want to make this change, it will be a larger change (addressed by a separate PR) since we will need to be consistent across all reports and APIs.

pvk-developer (approved):

PR #242, opened 2022-10-07
Update visualization average per table

katxiao:
Update visualization titles to use per-table averages instead of the overall average.


fealho (approved):
👍

PR #244, opened 2022-10-13
Accept both old and new metadata

katxiao:
Update single table metrics to accept both old and new metadata


sdmetrics/demos.py
```diff
@@ -15,14 +15,19 @@ def load_demo(modality='multi_table'):
...
     Returns:
         tuple:
             Real data, Synthetic data, Metadata.
     """
     demo_path = pathlib.Path(__file__).parent / 'demos' / f'{modality}.pkl'
     with open(demo_path, 'rb') as demo_file:
-        return pickle.load(demo_file)
+        last_index = 3
+        if new_metadata:
+            last_index = 4
```
amontanez24, line 29: why does the last index change? Can we add a comment or change a parameter name to indicate what this is for? Also minor but this can be written on one line as 
```python
last_index = 4 if new_metadata else 3
```

sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -70,16 +70,16 @@ def compute_breakdown(cls, real_data, synthetic_data, metadata=None,
...
+            field_type = cls._get_type_from_column_meta(field_meta)
+
+            if field_type == 'datetime':
                 real_data[field] = pd.to_numeric(real_data[field])
                 synthetic_data[field] = pd.to_numeric(synthetic_data[field])
                 numerical_fields.append(field)
-            elif field_meta['type'] == 'numerical':
+            elif field_type == 'numerical':
                 numerical_fields.append(field)
-            else:
```
amontanez24, line 81: why did we get rid of this case?

amontanez24 (commented):
This looks pretty good to me! Just a few minor comments/questions mostly for clarity

sdmetrics/demos.py
katxiao: The new metadata is stored as the 4th index, so depending on if we want to return it or not, the `last_index` changes. Sounds good, I'll add a comment

sdmetrics/single_table/new_row_synthesis.py
katxiao: The `discrete_fields` variable wasn't being used. In the code later on, we just assume everything not in `numerical_fields` is discrete.

amontanez24 (approved):

PR #248, opened 2022-10-18
Add default synthetic sample size to DiagnosticReport

katxiao:


sdmetrics/reports/multi_table/diagnostic_report.py
```diff
@@ -33,6 +33,10 @@ class DiagnosticReport():
         'Boundaries': [BoundaryAdherence],
     }
 
+    _METRIC_ARGS = {
```
amontanez24, line 36: Is there a reason to make this a class attribute instead of an instance attribute? I just wonder because if you change the values in one then I think the next `DiagnosticReport` created will also have the changed value

katxiao: Ah I see, I can move it to an instance attribute, good call

sdmetrics/reports/multi_table/diagnostic_report.py
```diff
@@ -76,12 +79,15 @@ def generate(self, real_data, synthetic_data, metadata):
                 The metadata, which contains each column's data type as well as relationships.
         """
         metrics = list(itertools.chain.from_iterable(self.METRICS.values()))
+        self._metric_args['NewRowSynthesis']['synthetic_sample_size'] = min(
+            len(real_data), self._metric_args['NewRowSynthesis']['synthetic_sample_size'])
 
         for metric in tqdm.tqdm(metrics, desc='Creating report'):
             metric_name = metric.__name__
             try:
+                args = self._metric_args.get(metric_name, {})
```
pvk-developer, line 88: The parameter `args` confuses me since it's a special key or in some cases it can be reserved for the method definition eg: `def func(*args, **kwargs)` .I would suggest to change `args` to something like `metric_args` or similar just to avoid this type of confusion.

pvk-developer (approved):
I left a suggestion on renaming the `args` to `metric_args` other than that it looks good to merge.

PR #251, opened 2022-10-24
Solving primary_key use for detection metrics

### LINKED ISSUE #119: Detection test should be aware of primary_key
### Environment details

If you are already running SDMetrics, please indicate the following details about the environment in
which you are running it:

* SDMetrics version: 2.4.2-dev0
* Python version: 3.9
* Operating System: ubuntu 20.04

### Problem description

When the primary_key is set, the generated data index restarts from zero.
As a consequence, detection metrics can trivially detect generated instances by setting a threshold on the primary_key.

### What I already tried

I will propose a patch to remove primary_key columns if sets form these tests.

R-Palazzo:
First pull request, Romain Palazzo for issue 119.
Resolves #119 
Thank you for your help :)


tests/unit/single_table/efficacy/detection/test_detection.py
```diff
@@ -0,0 +1,64 @@
+from unittest.mock import patch
+
+import pandas as pd
+
+from sdmetrics.single_table import LogisticDetection
+from tests.utils import DataFrameMatcher
+
+
+def test_primary_key_detection_metrics():
```
katxiao, line 9: Since this is testing a class method, let's make this test inside a class as well.

```
class TestDetectionMetric:

    def test_compute(self):
        ...
```

tests/unit/single_table/efficacy/detection/test_detection.py
```diff
@@ -0,0 +1,64 @@
+from unittest.mock import patch
```
katxiao, line 1: Since this is testing the `sdmetrics/single_table/detection/base.py` file, let's move this to a file at the location `tests/unit/single_table/detection/test_base.py`.

katxiao (changes requested):
Thanks @R-Palazzo! The code is looking good so far. I have a couple of comments around formatting and file structure.

Also could we update the name of the PR to something more descriptive of the PR? We don't have to include the issue number inside the PR name.

tests/unit/single_table/efficacy/detection/test_detection.py
R-Palazzo: Done, just I have an error when the filename is `test_base.py` so I let `test_detection.py`. Therefore it is `test/unit/signle_table/detection/test_detection.py` .

R-Palazzo (commented):
Done in @R-Palazzo@be18981.

katxiao (approved):
Awsome, LGTM!

PR #252, opened 2022-10-26
Move BNLikelihood to a subclass

### LINKED ISSUE #148: BNLikelihood not properly listed as a single table metric
### Environment Details

* SDMetrics version: 0.5.0
* Python version: 3.9
* Operating System: Windows

### Error Description

I want to go through a number of tests using single table metrics, but Bayesian Networks(BNLikelihood do not work using the method below, where metric is a list of tests I want to do:

tests=SingleTableMetric.get_subclasses()
out=tests[metric].compute(real_df, synth_df, metadata=config)

To add to this, in the readme here(https://github.com/sdv-dev/SDMetrics/tree/master/sdmetrics/single_table), BNLikelihood is not listed as output of SingleTableMetric.get_subclasses(), though it is mentioned in the readme as a test.

I do realise I can use BNLikelihood.compute() instead, so this probably isnt high priority, but I thought I should mention it.

### Steps to reproduce

See above

I dont have the time to reproduce the error right now, but the error message was a typical "this key does not exist" type of thing.

katxiao:
Resolves #148 

Currently, `BNLikelihood` is not returned in `SingleTableMetric.get_subclasses` because it is an intermediate child class. Move it to a terminal subclass so that it is considered a "final" implementation.

### DIFF
--- a/sdmetrics/single_table/bayesian_network.py
+++ b/sdmetrics/single_table/bayesian_network.py
@@ -11,29 +11,8 @@ from sdmetrics.single_table.base import SingleTableMetric
 LOGGER = logging.getLogger(__name__)
 
 
-class BNLikelihood(SingleTableMetric):
-    """BayesianNetwork Likelihood Single Table metric.
-
-    This metric fits a BayesianNetwork to the real data and then evaluates how
-    likely it is that the synthetic data belongs to the same distribution.
-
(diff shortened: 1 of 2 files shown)


pvk-developer (approved):
Should we add an integration tests that checks or ensures that this is now returned as a subclass ? 🤔

PR #259, opened 2022-11-10
Change columns to column_names

### LINKED ISSUE #258: Visualization `get_column_pair_plot`: update parameter name to `column_names`
### Environment Details

* SDMetrics version: 0.7 and 0.8
* Python version: Any
* Operating System: Any

### Error Description

The [docs](https://docs.sdv.dev/sdmetrics/reports/visualization-utilities#compare-a-pair-of-synthetic-columns-and-real-columns-2d) for `utils.get_column_pair_plot` mention that the column names should be given as a parameter `column_names`.  This yields an error. The method seems to expect `columns` as the parameter name instead.

### Expected Results 
We should change the parameter to be consistent with the docs:

(a) We use `columns_names(s)` whenever the user is providing a string with the names, instead of the actual column data
(b) The other visualization method `get_column_plot` has the desired `column_name` so it would be consistent.

katxiao:
Resolves #258

### DIFF
--- a/sdmetrics/reports/utils.py
+++ b/sdmetrics/reports/utils.py
@@ -375,7 +375,7 @@ def make_mixed_column_pair_plot(real_data, synthetic_data):
     return fig
 
 
-def get_column_pair_plot(real_data, synthetic_data, columns, metadata):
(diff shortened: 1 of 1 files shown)


amontanez24 (approved):
:shipit:

PR #261, opened 2022-11-15
Use base class method to get valid fields for type

### LINKED ISSUE #260: NewRowSynthesis should ignore PrimaryKey column
### Problem Description

It appears that the `NewRowSynthesis` measure includes the `primary_key` column in its analysis. I can feed it two tables that are identical except for the `primary_key` column, and it will measure every synthetic data row as new even though practically speaking none of them are new.

### Expected behavior

Ignore the `primary_key` column when doing `NewRowSynthesis`.

katxiao:
Resolves #260

### DIFF
--- a/sdmetrics/single_table/new_row_synthesis.py
+++ b/sdmetrics/single_table/new_row_synthesis.py
@@ -3,8 +3,10 @@ import warnings
 
 import pandas as pd
 
+from sdmetrics.errors import IncomputableMetricError
 from sdmetrics.goal import Goal
 from sdmetrics.single_table.base import SingleTableMetric
+from sdmetrics.utils import get_columns_from_metadata, get_type_from_column_meta
 
 
 class NewRowSynthesis(SingleTableMetric):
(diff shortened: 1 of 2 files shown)


sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -70,30 +71,36 @@ def compute_breakdown(cls, real_data, synthetic_data, metadata=None,
...
+            if get_type_from_column_meta(field_meta) == 'datetime':
                 real_data[field] = pd.to_numeric(real_data[field])
                 synthetic_data[field] = pd.to_numeric(synthetic_data[field])
-                numerical_fields.append(field)
-            elif field_type == 'numerical':
-                numerical_fields.append(field)
+
+        try:
+            numerical_fields = cls._select_fields(metadata, ('numerical', 'datetime'))
+        except IncomputableMetricError:
```
amontanez24, line 81: Kind of unrelated by what does this method throw an `IncomputableMetricError`? It doesn't seem like it's computing a metric but maybe I'm misunderstanding

sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -70,30 +71,36 @@ def compute_breakdown(cls, real_data, synthetic_data, metadata=None,
...
+            numerical_fields = []
+
+        try:
+            categorical_fields = cls._select_fields(metadata, ('categorical', 'boolean'))
+        except IncomputableMetricError:
+            categorical_fields = []
```
amontanez24, line 87: very minor: but you could wrap these try/excepts in a method that gets the fields of a type, catches the error and returns an empty list if it does.

amontanez24 (approved):
Seems good! Some very minor comments

sdmetrics/single_table/new_row_synthesis.py
katxiao: I think the original usage was like - if there are no fields of the valid type, we should throw an `IncomputableMetricError`. I do agree that doesn't make much sense to throw it in `_select_fields` - the method that is calling `_select_fields` should throw the error.

Issue #296, opened 2023-01-18
Synthetic data missing values should be multiplied by 100 (in `utils.get_column_plot`)


npatki:
### Environment Details

* SDMetrics version: 0.8.1 (latest)
* Python version: Any
* Operating System: Any

### Error Description

Use `utils.get_column_plot` to visualize a column that contains missing values. Observe that Missing Values %ages are printed at the top right: The original data is represented as a %age (between 0 and 100) but the synthetic data is only shown as a proportion (between 0 and 1). The synthetic data should be multiplied by 100.

![image](https://user-images.githubusercontent.com/3227804/213052388-556fe886-c609-4952-8f3f-cf323885a255.png)

In the example above, synthetic data should be 43%, not 0.43%.

### Steps to reproduce

Create synthetic data using the `student_placements_pii` dataset and visualize the column `start_date`.

```python
from sdv.demo import load_tabular_demo
from sdv.tabular import GaussianCopula
from sdmetrics.reports.utils import get_column_plot

metadata, data = load_tabular_demo('student_placements_pii', metadata=True)
synthesizer = GaussianCopula(table_metadata=metadata)
synthesizer.fit(data)

synthetic_data = synthesizer.sample(num_rows=len(data))

# this prints out 31% for real data and 43% for synthetic data
print('Missing values real data', len(data[data['start_date'].isnull()])/len(data) * 100, '%')
print('Missing values synthetic data', len(synthetic_data[synthetic_data['start_date'].isnull()])/len(synthetic_data) * 100, '%')

# this shows 31% for real data and 0.43% for synthetic data
fig = get_column_plot(data, synthetic_data, 'start_date', metadata.to_dict())
fig.show()
```

npatki:
Resolved in 0.9.1

PR #304, opened 2023-02-09
fix issue-296 for discrete and continuous columns

R-Palazzo:


R-Palazzo:
Resolves #296

sdmetrics/reports/utils.py
```diff
@@ -86,7 +86,8 @@ def make_discrete_column_plot(real_column, synthetic_column, sdtype):
     synthetic_data['Data'] = 'Synthetic'
 
     missing_data_real = round((real_column.isna().sum() / len(real_column)) * 100, 2)
-    missing_data_synthetic = round((synthetic_column.isna().sum() / len(synthetic_column)), 2)
+    missing_data_synthetic = round(
```
amontanez24, line 89: I think one way we can avoid bugs like this in the future is by pulling all the logic into on common placce. I think it makes sense to have a private static method `_get_missing_percentage` that takes in the column and returns the percentage and then we can just call it on lines 88, 89, 157 and 158

sdmetrics/reports/utils.py
```diff
@@ -154,7 +155,8 @@ def make_continuous_column_plot(real_column, synthetic_column, sdtype):
     """
     column_name = real_column.name if hasattr(real_column, 'name') else ''
     missing_data_real = round((real_column.isna().sum() / len(real_column)) * 100, 2)
-    missing_data_synthetic = round((synthetic_column.isna().sum() / len(synthetic_column)), 2)
+    missing_data_synthetic = round(
```
amontanez24, line 158: Can we also add a check in the unit tests to make sure this is correct now?

amontanez24 (changes requested):

sdmetrics/reports/utils.py
R-Palazzo: Yes I can

sdmetrics/reports/utils.py
R-Palazzo: Yes, it's a good idea, I can do this

sdmetrics/reports/utils.py
```diff
@@ -153,8 +154,8 @@ def make_continuous_column_plot(real_column, synthetic_column, sdtype):
         plotly.graph_objects._figure.Figure
     """
     column_name = real_column.name if hasattr(real_column, 'name') else ''
-    missing_data_real = round((real_column.isna().sum() / len(real_column)) * 100, 2)
-    missing_data_synthetic = round((synthetic_column.isna().sum() / len(synthetic_column)), 2)
+    missing_data_real = round(get_missing_percentage(real_column), 2)
```
amontanez24, line 157: I think even the rounding part could go in the `get_missing_percentage` function

amontanez24 (approved):
One small comment but after that :shipit:

sdmetrics/reports/utils.py
R-Palazzo: Moved the rounding part inside the function.

amontanez24 (approved):
:shipit:

PR #313, opened 2023-02-24
Fix `ValueError` in `NewRowSynthesis` if size of `row_filter` exceeds 31

### LINKED ISSUE #307: Fix `ValueError` in the `NewRowSynthesis` metric (also impacts `DiagnosticReport`)
### Environment Details

* SDMetrics version: 0.9.0
* Python version: 3.8

### Error Description
If my data contains >32 columns, then the `NewRowSynthesis` metric crashes with a `ValueError`. This means that I can't run the `DiagnosticReport`. 

### Workaround
Until we make the fix, you can work around this issue by _choosing a smaller set of columns_ when running this metric or a diagnostic report. As long as there are 32 or fewer columns, it seems to work.

### Technical Fix
If there is an error in the [pandas query method](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html), we should try calling it again with `engine='python'`.

In [this line of code](https://github.com/sdv-dev/SDMetrics/blob/master/sdmetrics/single_table/new_row_synthesis.py#L113) we are only excepting a `TypeError`. We should also except a `ValueError`. Something like the following:

```python
try:
    matches = real_data.query(' and '.join(row_filter))
except TypeError:
    if len(real_data) > 10000:
        warnings.warn('Unable to optimize query. For better formance, set the `synthetic_sample_size` parameter or upgrade to Python 3.8')
    matches = real_data.query(' and '.join(row_filter), engine='python')
except ValueError:
    matches = real_data.query(' and '.join(row_filter), engine='python')
```

This seems to work when I try it!

frances-h:
Resolve #307 

Set the query engine to `'python'` if more than 31 `row_filter`s are specified. Trying to use the default `numexpr` will result in a `ValueError` or `Memory` error if more than 31 `row_filter`s are provided, probably due to `numpy` limiting `NPY_MAXARGS=32`.

### DIFF
--- a/sdmetrics/single_table/new_row_synthesis.py
+++ b/sdmetrics/single_table/new_row_synthesis.py
@@ -108,9 +108,12 @@ class NewRowSynthesis(SingleTableMetric):
 
                 row_filter.append(field_filter)
 
+            engine = None
+            if len(row_filter) >= 32:  # Limit set by NPY_MAXARGS
+                engine = 'python'
             try:
-                matches = real_data.query(' and '.join(row_filter))
-            except TypeError:
+                matches = real_data.query(' and '.join(row_filter), engine=engine)
+            except (TypeError, ValueError):
                 if len(real_data) > 10000:
                     warnings.warn('Unable to optimize query. For better formance, set the '
                                   '`synthetic_sample_size` parameter or upgrade to Python 3.8')
--- a/sdmetrics/single_table/new_row_synthesis.py
+++ b/sdmetrics/single_table/new_row_synthesis.py
@@ -113,7 +113,7 @@ class NewRowSynthesis(SingleTableMetric):
                 engine = 'python'
             try:
                 matches = real_data.query(' and '.join(row_filter), engine=engine)
-            except (TypeError, ValueError):
+            except TypeError:
                 if len(real_data) > 10000:
                     warnings.warn('Unable to optimize query. For better formance, set the '
                                   '`synthetic_sample_size` parameter or upgrade to Python 3.8')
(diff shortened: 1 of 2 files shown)


sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -108,9 +108,12 @@ def compute_breakdown(cls, real_data, synthetic_data, metadata=None,
...
+            engine = None
+            if len(row_filter) >= 32:  # Limit set by NPY_MAXARGS
+                engine = 'python'
             try:
-                matches = real_data.query(' and '.join(row_filter))
-            except TypeError:
+                matches = real_data.query(' and '.join(row_filter), engine=engine)
+            except (TypeError, ValueError):
                 if len(real_data) > 10000:
                     warnings.warn('Unable to optimize query. For better formance, set the '
```
npatki, line 118: @frances-h if you're running into the `ValueError` due to >32 columns, then we don't need to show the warning.

I think the warning is only for `TypeError`, whose root cause is [a different issue](https://github.com/pandas-dev/pandas/issues/22435).

frances-h: Good point, we actually don't need to catch the `ValueError` anymore since we won't try to use `numexpr` if there's too many arguments.

tests/unit/single_table/test_new_row_synthesis.py
```diff
@@ -144,6 +144,32 @@ def test_compute_with_sample_size_too_large(self, warnings_mock):
             'synthetic data rows (5). Proceeding without sampling.'
         )
 
+    def test_compute_with_many_columns(self):
+        """Test the ``compute`` method with more than 32 columns.
+
+           Expect that the new row synthesis is returned.
```
pvk-developer, line 150: Fix indent.

pvk-developer (approved):
LGTM!

fealho (approved):
👍

PR #323, opened 2023-03-15
Better extra categoricals warning message

### LINKED ISSUE #315: More detailed warning in `QualityReport` for unexpected category values
### Problem Description
Currently when you run the `QualityReport`, the software checks to see whether all the category _values_ in the synthetic data were present in the real data. This is a good check, as categorical columns must have a predefined set of values -- it doesn't make sense for the synthetic data to have unseen categories.

However, the warning makes it hard to identify where exactly the unseen categories are coming from. Here's what's shown today:

```
UserWarning: Unexpected value <VALUE> in synthetic data.
```

This is not useful, especially if it prints out many times.
```
UserWarning: Unexpected value A in synthetic data.
UserWarning: Unexpected value B in synthetic data.
UserWarning: Unexpected value C in synthetic data.
...
```

### Expected behavior

**For multi-table data**: For a multi table `QualityReport`, we should print table and column name. We can squash all the category values into a single warning per column.

```
UserWarning: Unexpected values ('A', 'B', 'C') in column 'gender' and table 'users'
UserWarning: Unexpected values ('X', 'Y', 'Z', 'AA', 'BB' + more) in column 'status' and table 'sessions'
```

**For single-table data**: For a single table `QualityReport`, we do not have to print the table name
```
UserWarning: Unexpected values ('A', 'B', 'C') in column 'gender'
UserWarning: Unexpected values ('X', 'Y', 'Z', 'AA', 'BB' + more) in column 'status'
```

frances-h:
Resolve #315

### DIFF
--- a/sdmetrics/reports/utils.py
+++ b/sdmetrics/reports/utils.py
@@ -641,8 +641,8 @@ def _validate_categorical_values(real_data, synthetic_data, metadata, table=None
             The name of the current table, if one exists
     """
     if table:
-        warning_format = 'Unexpected values ({values}) in column "{column}" '\
-                         f'and table "{table}"'
+        warning_format = ('Unexpected values ({values}) in column "{column}" '
+                          f'and table "{table}"')
     else:
         warning_format = 'Unexpected values ({values}) in column "{column}"'
 
--- a/sdmetrics/reports/utils.py
+++ b/sdmetrics/reports/utils.py
@@ -656,7 +656,8 @@ def _validate_categorical_values(real_data, synthetic_data, metadata, table=None
             ]
             if extra_categories:
                 value_list = '", "'.join(str(value) for value in extra_categories[:5])
-                values = f'"{value_list}" + more' if len(extra_categories) > 5 else f'"{value_list}"'
+                values = f'"{value_list}" + more' if len(
+                    extra_categories) > 5 else f'"{value_list}"'
                 warnings.warn(warning_format.format(values=values, column=column))
 
 
--- a/sdmetrics/reports/utils.py
+++ b/sdmetrics/reports/utils.py
@@ -655,12 +655,8 @@ def _validate_categorical_values(real_data, synthetic_data, metadata, table=None
                 if value not in real_data[column].unique()
             ]
             if extra_categories:
-                if len(extra_categories) > 5:
-                    value_list = '", "'.join(str(value) for value in extra_categories[:5])
-                    values = f'"{value_list}" + more'
-                else:
-                    value_list = '", "'.join(str(value) for value in extra_categories)
-                    values = f'"{value_list}"'
+                value_list = '", "'.join(str(value) for value in extra_categories[:5])
+                values = f'"{value_list}" + more' if len(extra_categories) > 5 else f'"{value_list}"'
                 warnings.warn(warning_format.format(values=values, column=column))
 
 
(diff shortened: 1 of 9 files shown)


sdmetrics/reports/utils.py
```diff
@@ -624,3 +625,79 @@ def print_results_for_level(out, results, level):
...
+                if value not in real_data[column].unique()
+            ]
+            if extra_categories:
+                if len(extra_categories) > 5:
+                    value_list = '", "'.join(str(value) for value in extra_categories[:5])
+                    values = f'"{value_list}" + more'
+                else:
+                    value_list = '", "'.join(str(value) for value in extra_categories)
+                    values = f'"{value_list}"'
+                warnings.warn(warning_format.format(values=values, column=column))
```
amontanez24, line 664: minor: add a space above this line

sdmetrics/reports/utils.py
```diff
@@ -624,3 +625,79 @@ def print_results_for_level(out, results, level):
...
+            extra_categories = [
+                value for value in synthetic_data[column].unique()
+                if value not in real_data[column].unique()
+            ]
+            if extra_categories:
+                if len(extra_categories) > 5:
+                    value_list = '", "'.join(str(value) for value in extra_categories[:5])
+                    values = f'"{value_list}" + more'
+                else:
+                    value_list = '", "'.join(str(value) for value in extra_categories)
```
amontanez24, line 662: minor: we could shorten this to
```python
value_list = '", "'.join(str(value) for value in extra_categories[:5])
values = f'"{value_list}" + more' if len(extra_categories) > 5 else f'"{value_list}"'
```
since taking a subset of a list with less elements than the max index doesn't crash, but the current way is maybe more explicit

tests/unit/reports/test_utils.py
```diff
@@ -1189,3 +1191,127 @@ def test_aggregate_metric_results():
...
+
+
+def test__validate_categorical_values():
+    """Test no extra categoricals does not crash."""
+    # Setup
+    sdtype = 'categorical'
+    real_data = pd.DataFrame({'col1': [1, 2, 3, 4]})
+    synthetic_data = pd.DataFrame({'col1': [1, 2, 4, 4]})
+    metadata = {'fields': {'col1': {'type': sdtype}}}
+    warnings.filterwarnings('error', category=UserWarning)
```
amontanez24, line 1203: do we need to filter and reset warnings for this one? My understanding is that none are getting raised here and it doesn't look like we're checking them

amontanez24 (approved):
Some minor comments but this LGTM!

tests/unit/reports/test_utils.py
frances-h: Yeah, my intention here was to test that the warning doesn't get raised when there aren't extra categories. Since it's just a warning, the only way to get the test to fail if the warning occurs is to have the warning to error.

sdmetrics/reports/utils.py
```diff
@@ -624,3 +625,75 @@ def print_results_for_level(out, results, level):
...
+        synthetic_data (pd.DataFrame):
+            The synthetic data.
+        metadata (dict):
+            The metadata.
+        table (str, optional):
+            The name of the current table, if one exists
+    """
+    if table:
+        warning_format = 'Unexpected values ({values}) in column "{column}" '\
+                         f'and table "{table}"'
```
pvk-developer, line 645: In general when having to deal with this situations I would recommend using `(`. Also looks like missing formatting for `values` and `column` (f-string).

pvk-developer (approved):
Just a minor comment but other than that LGTM!

PR #326, opened 2023-03-21
Add parent/child dtype validation

### LINKED ISSUE #317: Make error more informative in `QualityReport` when tables cannot be merged
### Problem Description
If I try to run the `QualityReport` on multi table data, then the connections between the tables must be valid. It should be possible to merge the parent and child tables using the primary and foreign key connections.

If the merge fails (for eg. the foreign key is a different data type than the primary key), then an error is shown:
```
ValueError: You are trying to merge on int64 and object columns. If you wish to proceed you should use pd.concat
```

This error comes from `pandas` and is not very helpful to diagnose what's wrong. 

### Expected behavior

We should throw a more descriptive error message that tells the user which exact table(s) and column(s) have mismatch in data.

```
Error: The 'user' and 'session' table cannot be merged. Please make sure the primary key in 'user' ('id') and
the foreign key in 'session' ('user_id') have the same data type.
```

frances-h:
Resolve #317

### DIFF
--- a/tests/unit/reports/test_utils.py
+++ b/tests/unit/reports/test_utils.py
@@ -1346,7 +1346,7 @@ def test_validate_multi_table_parent_child_dtype_mismatch():
     error_msg = re.escape(
         "The 'table1' table and 'table2' table cannot be merged. Please make sure the primary key "
         "in 'table1' ('primary_key') and the foreign key in 'table2' ('foreign_key') have the same"
-        " data type."
+        ' data type.'
     )
 
     # Run and Assert


sdmetrics/reports/utils.py
```diff
@@ -682,6 +682,20 @@ def validate_multi_table_inputs(real_data, synthetic_data, metadata):
...
+        child_dtype = real_data[rel['child_table_name']][rel['child_foreign_key']].dtype
+        if (parent_dtype == 'object' and child_dtype != 'object') or (
+                parent_dtype != 'object' and child_dtype == 'object'):
+            parent = rel['parent_table_name']
+            parent_key = rel['parent_primary_key']
+            child = rel['child_table_name']
+            child_key = rel['child_foreign_key']
+            error_msg = (f'The "{parent}" table and "{child}" table cannot be merged. Please '
+                         f'make sure the primary key in "{parent}" ("{parent_key}") and the '
+                         f'foreign key in "{child}" ("{child_key}") have the same data type.')
```
pvk-developer, line 696: Let's use `'` rather than `"` to high light the parent and child table names to keep consistency across the libraries.

PR #342, opened 2023-04-28
Make DiagnosticReport more fault tolerant

### LINKED ISSUE #332: Make the diagnostic report more fault-tolerant
### Problem Description
As described in issue #327, the `DiagnosticReport` currently crashes during generation if the `NewRowSynthesis` metric fails.

### Expected Behavior
Regardless of the actual reason why a metric is failing, we expect the `DiagnosticReport` to catch that error and show the score as `NaN`. This way, the user will still be able to see the _other metrics_ that have been computed as part of the diagnostic.

This applies to all metrics / properties that are computed in the `DiagnosticReport` (Synthesis, Coverage, Boundaries). See [docs](https://docs.sdv.dev/sdmetrics/reports/diagnostic-report/whats-included) for more details.

frances-h:
Resolve #332

### DIFF
--- a/sdmetrics/reports/multi_table/diagnostic_report.py
+++ b/sdmetrics/reports/multi_table/diagnostic_report.py
@@ -2,6 +2,7 @@
 
 import copy
 import itertools
+import logging
 import pickle
 import sys
 import warnings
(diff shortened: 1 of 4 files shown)


fealho (commented):
Looks good. Could you just test the `nan`s are being displayed when the metric fails?

amontanez24 (approved):
LGTM!

fealho (approved):

PR #343, opened 2023-05-02
Issue 327 newrowsynthesis special characters in column name.

### LINKED ISSUE #327: ValueError: multi-line expressions (`NewRowSynthesis` metric in `DiagnosticReport`)
### Environment Details

* SDMetrics version: 0.9.2 (latest) and previous versions (0.9.0, 0.8.0)
* Python version: 3.7

### Error Description
The `DiagnosticReport` crashes on the `NewRowSynthesis` metric due to a multi-line expression in the query method. 

```python
report = DiagnosticReport()
report.generate(real_data, synthetic_data, metadata)
```

```
ValueError: multi-line expressions are only valid in the context of data, use DataFrame.eval
```

### Replication
I'm unable to share the exact data and metadata (as it's a private dataset). But here are the following key statistics:

- 178 total columns in a single table
- 153 categorical and no boolean columns
- 9 datetime and 1 numerical column
- 1 id column
- 14 pii columns

### Stack Trace

```
--------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[<ipython-input-7-4c304e411f24>](https://localhost:8080/#) in <module>
      5 
      6 start = time.time()
----> 7 d_report.generate(
      8     real_data=dev_data,
      9     synthetic_data=dev_data_output,

4 frames
[/usr/local/lib/python3.9/dist-packages/sdmetrics/reports/single_table/diagnostic_report.py](https://localhost:8080/#) in generate(self, real_data, synthetic_data, metadata)
     89             try:
     90                 metric_args = self._metric_args.get(metric_name, {})
---> 91                 self._metric_results[metric_name] = metric.compute_breakdown(
     92                     real_data, synthetic_data, metadata, **metric_args)
     93 

[/usr/local/lib/python3.9/dist-packages/sdmetrics/single_table/new_row_synthesis.py](https://localhost:8080/#) in compute_breakdown(cls, real_data, synthetic_data, metadata, numerical_match_tolerance, synthetic_sample_size)
    103 
    104             try:
--> 105                 matches = real_data.query(' and '.join(row_filter))
    106             except TypeError:
    107                 if len(real_data) > 10000:

[/usr/local/lib/python3.9/dist-packages/pandas/core/frame.py](https://localhost:8080/#) in query(self, expr, inplace, **kwargs)
   4109         kwargs["level"] = kwargs.pop("level", 0) + 1
   4110         kwargs["target"] = None
-> 4111         res = self.eval(expr, **kwargs)
   4112 
   4113         try:

[/usr/local/lib/python3.9/dist-packages/pandas/core/frame.py](https://localhost:8080/#) in eval(self, expr, inplace, **kwargs)
   4238         kwargs["resolvers"] = tuple(kwargs.get("resolvers", ())) + resolvers
   4239 
-> 4240         return _eval(expr, inplace=inplace, **kwargs)
   4241 
   4242     def select_dtypes(self, include=None, exclude=None) -> DataFrame:

[/usr/local/lib/python3.9/dist-packages/pandas/core/computation/eval.py](https://localhost:8080/#) in eval(expr, parser, engine, truediv, local_dict, global_dict, resolvers, level, target, inplace)
    324 
    325     if multi_line and target is None:
--> 326         raise ValueError(
    327             "multi-line expressions are only valid in the "
    328             "context of data, use DataFrame.eval"

ValueError: multi-line expressions are only valid in the context of data, use DataFrame.eval
```

R-Palazzo:
Resolve #327.

### DIFF
--- a/sdmetrics/single_table/new_row_synthesis.py
+++ b/sdmetrics/single_table/new_row_synthesis.py
@@ -86,12 +86,24 @@ class NewRowSynthesis(SingleTableMetric):
         except IncomputableMetricError:
             categorical_fields = []
 
+        def update_column_name(char, column_name):
+            return column_name.replace(char, '_') if char in column_name else column_name
+
+        def update_lists(column, new_column):
+            for field_list in [numerical_fields, categorical_fields]:
+                if column in field_list:
+                    field_list.remove(column)
+                    field_list.append(new_column)
+
         for column in real_data.columns:
-            if '\n' in column:
-                real_data = real_data.rename(columns={column: column.replace('\n', ' ')})
-                synthetic_data = synthetic_data.rename(columns={column: column.replace('\n', ' ')})
-        real_data.columns = real_data.columns.str.replace(r"[.']+", '_', regex=True)
-        synthetic_data.columns = synthetic_data.columns.str.replace(r"[.']", '_', regex=True)
+            new_column = column
+            for char in ['\n', '.', "'"]:
+                new_column = update_column_name(char, new_column)
+
+            if new_column != column:
+                real_data = real_data.rename(columns={column: new_column})
+                synthetic_data = synthetic_data.rename(columns={column: new_column})
+                update_lists(column, new_column)
 
         num_unique_rows = 0
         for index, row in synthetic_data.iterrows():
@@ -114,7 +126,6 @@ class NewRowSynthesis(SingleTableMetric):
                         field_filter = f'`{field}` == {row[field]}'
 
                 row_filter.append(field_filter)
-
             engine = None
             if len(row_filter) >= 32:  # Limit set by NPY_MAXARGS
                 engine = 'python'
--- a/sdmetrics/single_table/new_row_synthesis.py
+++ b/sdmetrics/single_table/new_row_synthesis.py
@@ -86,6 +86,13 @@ class NewRowSynthesis(SingleTableMetric):
         except IncomputableMetricError:
             categorical_fields = []
 
+        for column in real_data.columns:
+            if '\n' in column:
+                real_data = real_data.rename(columns={column: column.replace('\n', ' ')})
+                synthetic_data = synthetic_data.rename(columns={column: column.replace('\n', ' ')})
+        real_data.columns = real_data.columns.str.replace("'", "_", regex=True).str.replace(".", "", regex=True)
+        synthetic_data.columns = synthetic_data.columns.str.replace("'", "_", regex=True).str.replace(".", "", regex=True)
+
         num_unique_rows = 0
         for index, row in synthetic_data.iterrows():
             row_filter = []
@@ -112,7 +119,7 @@ class NewRowSynthesis(SingleTableMetric):
             if len(row_filter) >= 32:  # Limit set by NPY_MAXARGS
                 engine = 'python'
             try:
-                matches = real_data.query(' and '.join(row_filter), engine=engine)
+                matches = real_data.query(' and '.join(row_filter))
             except TypeError:
                 if len(real_data) > 10000:
                     warnings.warn('Unable to optimize query. For better formance, set the '
--- a/sdmetrics/single_table/new_row_synthesis.py
+++ b/sdmetrics/single_table/new_row_synthesis.py
@@ -126,6 +126,7 @@ class NewRowSynthesis(SingleTableMetric):
                         field_filter = f'`{field}` == {row[field]}'
 
                 row_filter.append(field_filter)
+
             engine = None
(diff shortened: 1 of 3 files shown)


fealho (commented):
Could you add a test case which used to fail but doesn't anymore after the fix?

tests/integration/single_table/test_single_table.py
```diff
@@ -123,3 +124,27 @@ def test_compute_all():
...
+        'd.': [7, 8, 9],
+        "e.'\n": [10, 11, 12],
+    })
+
+    # Run
+    metric = NewRowSynthesis.compute(
+        real_data=real_data,
+        synthetic_data=synthetic_data
+    )
+
```
fealho, line 149: Add # Assert

tests/integration/single_table/test_single_table.py
```diff
@@ -123,3 +124,27 @@ def test_compute_all():
 
     scores = output[output.normalized_score.notna()]
     assert scores.normalized_score.between(0.0, 1.0).all()
+
+
+def test_newrowsynthesis_with_special_characters_in_column_names():
+    # Setup
```
fealho, line 130: Add small docstring description.

fealho (approved):

sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -86,6 +86,25 @@ def compute_breakdown(cls, real_data, synthetic_data, metadata=None,
...
+
+        def update_lists(column, new_column):
+            for field_list in [numerical_fields, categorical_fields]:
+                if column in field_list:
+                    field_list.remove(column)
+                    field_list.append(new_column)
+
+        for column in real_data.columns:
+            new_column = column
+            for char in ['\n', '.', "'"]:
```
amontanez24, line 100: Can we change the `update_column_name` method to take in a list of characters to replace and loop on its own? And can we rename it to `strip_characters` or something like that and just make it a utility function?

sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -86,6 +86,25 @@ def compute_breakdown(cls, real_data, synthetic_data, metadata=None,
         except IncomputableMetricError:
             categorical_fields = []
 
+        def update_column_name(char, column_name):
+            return column_name.replace(char, '_') if char in column_name else column_name
+
+        def update_lists(column, new_column):
```
amontanez24, line 92: since this method is nested and only used here, I think we can just call the for loop directly and get rid of it

sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -86,6 +86,25 @@ def compute_breakdown(cls, real_data, synthetic_data, metadata=None,
         except IncomputableMetricError:
             categorical_fields = []
 
+        def update_column_name(char, column_name):
+            return column_name.replace(char, '_') if char in column_name else column_name
```
amontanez24, line 90: Instead of replacing with an `_` is there a way to escape these characters?

sdmetrics/utils.py
```diff
@@ -275,3 +275,24 @@ def get_alternate_keys(metadata):
...
+            The column name.
+
+    Returns:
+        string:
+            The column name with the characters stripped.
+    """
+    result = column_name
+    for character in list_character:
+        if character in result:
+            result = result.replace(character, '')
```
amontanez24, line 296: Does it work if you escape the characters instead of dropping them? As in adding a `\` before the character

R-Palazzo: No I tried doing
`result = result.replace(character, '\\' + character)`
which adds a single `\` before my character but now my test is failing with the initial error:
` ValueError: multi-line expressions are only valid in the context of data, use DataFrame.eval`

amontanez24: My only concern is that what if after stripping the characters from the column name, we end up with another column name that is already in the data. It's unlikely but possible right?

sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -86,6 +86,21 @@ def compute_breakdown(cls, real_data, synthetic_data, metadata=None,
         except IncomputableMetricError:
             categorical_fields = []
 
+        for column_name in real_data.columns:
+            new_column_name = strip_characters(['\n', '.', "'"], column_name)
+            if new_column_name != column_name:
+                if new_column_name in real_data.columns:
+                    while new_column_name in real_data.columns:
+                        new_column_name += '_'
```
amontanez24, line 94: minor: can you add a new line under this one?

amontanez24 (approved):
I think this is ready to go now! 🍾

PR #350, opened 2023-05-17
Uncap torch

### LINKED ISSUE #347: Upgrade to torch 2.0
### Problem Description
The SDMetrics library has an optional dependency on `torch`, which we have not yet upgraded to 2..0

```
pip install sdmetrics[torch]
```

Current requirements:
```
torch_requires = [
    "torch>=1.8.0,<2;python_version<'3.10'",
    "torch>=1.11.0,<2;python_version>='3.10'",
]
```

### Additional context
These older versions of torch are blocking us from upgrading to Python 3.11

(See issue #346)

fealho:
Resolve #347.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,8 +29,8 @@ pomegranate_requires = [
 ]
 
 torch_requires = [
-    "torch>=1.8.0,<2;python_version<'3.10'",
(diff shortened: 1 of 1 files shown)


frances-h (approved):
Glad it was that easy 👍

PR #367, opened 2023-06-13
Add `BaseMultiTable` property

### LINKED ISSUE #357: Create multi table BaseProperty class
### Problem Description

_As a developer, it would be easier to make changes to the `QualityReport` if our code abstractions matched our logical and user facing abstractions. It's hard to add new properties, modify aggregation in existing ones and handle errors now because the results are being collected in a way that doesn't match the desired output._
 
After adding the `BaseSingleTableProperty` class in #354, we can use it at the multi table level as well. To do this we can create a `BaseMultiTableProperty` class and add it to a `_properties` module in the `reports/multi_table` folder.

#### Attributes

- `metrics`: A list of metrics that make up the property.
- `_properties`: A dict mapping the table names to their single table properties.

#### Abstract methods

- `get_score` - Returns a float that is the average score of all the individual metric scores computed. This method should average all the scores for each column/column pair as opposed to getting the average score for each table and averaging those.
- `get_visualization(table_name)` - Returns a `plotly.graph_objects._figure.Figure` object for the specified table.

### Additional context

- In the case of column shape metrics, the score should be computed by averaging the scores of every column in every table. This way no table scores are weighted disproportionately. For column pair metrics, the same should happen except for every column pair for each table.

pvk-developer:
Resolves #357

### DIFF
--- a/sdmetrics/reports/multi_table/_properties/base.py
+++ b/sdmetrics/reports/multi_table/_properties/base.py
@@ -11,6 +11,7 @@ class BaseMultiTableProperty():
         properties (dict):
             A dict mapping the table names to their single table properties.
     """
+
     _properties = None
 
     def get_score(self, real_data, synthetic_data, metadata, progress_bar):
--- a/sdmetrics/reports/multi_table/_properties/base.py
+++ b/sdmetrics/reports/multi_table/_properties/base.py
@@ -1,0 +1,46 @@
+"""Multi table base property class."""
+
+
+class BaseMultiTableProperty():
+    """Base class for multi table properties.
+
+    A property is a higher-level concept for a class that loops through all the base-level data
+    and applies different base-level metrics based on the data type.
+
+    Attributes:
+        metrics (list):
+            A list of metrics that make up the property.
+        properties (dict):
+            A dict mapping the table names to their single table properties.
+    """
+
+    metrics = None
+    _properties = None
+
+    def get_score(self, real_data, synthetic_data, metadata, progress_bar):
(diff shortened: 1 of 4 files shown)


tests/unit/reports/multi_table/_properties/test_base.py
```diff
@@ -0,0 +1,31 @@
...
+    # Setup
+    base_property = BaseMultiTableProperty()
+
+    # Run and Assert
+    with pytest.raises(NotImplementedError):
+        base_property.get_score(None, None, None, None)
+
+
+def test_get_visualization():
+    """Test that the method returns the propertie's ``get_visualization``."""
```
fealho, line 21: typo: properties'

fealho (approved):

tests/unit/reports/multi_table/_properties/test_base.py
amontanez24: property's

sdmetrics/reports/multi_table/_properties/base.py
```diff
@@ -0,0 +1,46 @@
...
+    and applies different base-level metrics based on the data type.
+
+    Attributes:
+        metrics (list):
+            A list of metrics that make up the property.
+        properties (dict):
+            A dict mapping the table names to their single table properties.
+    """
+
+    metrics = None
```
amontanez24, line 17: I think we actually might not need this anymore

amontanez24 (approved):
Just some small comments but otherwise looks good

PR #368, opened 2023-06-20
Create single table column pair trends property

### LINKED ISSUE #356: Create single table column pair trends property
### Problem Description

Once #354 is complete we should migrate the code for the computation of property metrics to their respective classes. This issue is for the `ColumnPairTrends` property.

### Expected behavior

Add a `ColumnPairTrends` class to the `_properties` module. The class should inherit from the `BaseSingleTableProperty` class.

#### Attributes

- `metrics`: The metrics for this property are: [CorrelationSimilarity, ContingencySimilarity]
- `_details`: A dataframe containing the following columns
    - 'Table': Then table names
    - 'Column 1': The names of the first column involved
    - 'Column 2': The names of the second column involved
    - 'Metric': The name of the metric
    - 'Quality Score': The scores
    - 'Real Correlation': The score for the real data's correlation for the columns
    - 'Synthetic Correlation': The score for the synthetic data's correlation for the columns

#### Abstract methods

- `get_score(real_data, synthetic_data, metadata, progress_bar)` - Returns a float that is the average score of all the individual metric scores computed.
    - Pseudo code
    ```python
    discrete_real, discrete_synthetic = discretize(real_data, synthetic_data, metadata)
    n_unique_real = real_data.nunique()
    n_unique_synthetic = synthetic_data.nunique()

    # IMPORTANT: Keep in mind that correlation between column1, column2 is the same as 
    # the correlation between column2, column1 -- so we do not need to compute twice
    # if there are n columns, total number of computations are n(n+1)/2
    for column1, column2 in tqdm(all column pairs, disable=(not verbose)):
        try:
            if column1 == column2:
                pair_score = 1 # correlation between a column and itself is 1

            elif column1 or column 2 has < 2 unique values:
                pair_score = NaN # must have 2 or more values for a correlation

            elif both columns are continuous:
                pair_score = CorrelationSimilarity

            elif both columns are discrete:
                pair_score = ContingencySimilarity

            elif one is discrete and one is continuous:
                pair_score = ContingencySimilarity (discrete data)

            else:
                continue # don't compute correlation for PII or id

        except Exception as e:
            pair_score = NaN
            Warning("Unable to compute Column Pair trends for <column1> and <column2." + 
                    "Encountered Error: type(e).__name__, e"

    try:
        overall_property_score = average(all pair scores)
    except:
        overall_property_score = NaN
    ```
- `get_visualization(table_name)` - Returns a `plotly.graph_objects._figure.Figure` object for the specified table. Use code similar to what's in `get_column_pairs_plot`.

### Additional context

- See #354 for more info

R-Palazzo:
Resolve #356.

I made few changes outside of the Issue. Mainly:
- To drop the NaN only for the 2 columns on the `ColumnPairTrend` metrics, not over all the dataset.
- Changed bins definition when discretizing continuous columns to include more values. Bins are only defined on real_data, so discretization produces NaN if synthetic data have different values (higher or lower).


npatki:
@R-Palazzo, @amontanez24 

> Changed bins definition when discretizing continuous columns to include more values. Bins are only defined on real_data, so discretization produces NaN if synthetic data have different values (higher or lower).

This is the expected behavior:

1. Generate the bins based on the real data. For example, assume there are 10 bins that are labeled `0, 1, 2, .. 9`
2. Adjust the lowest and highest bin edges to be -/+ infinity. That is, bin `0` should be `[-inf, <value>]` and bin `9` shoudl be `[value, +inf]`
3. Now the bin values to discretize the synthetic data. There shouldn't be any nans due to the step before.
    - If a synthetic value is below the min of real data, it will be assigned to bin `0`
    - If a synthetic value is above the max of real data, it will be assigned to bin `9`

@R-Palazzo if there are any other changes you're making outside the issue, please Slack me just in case.

R-Palazzo:
> @R-Palazzo, @amontanez24
> 
> > Changed bins definition when discretizing continuous columns to include more values. Bins are only defined on real_data, so discretization produces NaN if synthetic data have different values (higher or lower).
> 
> This is the expected behavior:
> 
> 1. Generate the bins based on the real data. For example, assume there are 10 bins that are labeled `0, 1, 2, .. 9`
> 2. Adjust the lowest and highest bin edges to be -/+ infinity. That is, bin `0` should be `[-inf, <value>]` and bin `9` shoudl be `[<value>, +inf]`
> 3. Now use the bin edges to discretize the synthetic data. There shouldn't be any nans due to the step before.
>    
>    * If a synthetic value is below the min of real data, it will be assigned to bin `0`
>    * If a synthetic value is above the max of real data, it will be assigned to bin `9`
> 
> @R-Palazzo if there are any other changes you're making outside the issue, please Slack me just in case.

Yes thanks for your message. I didn't precise in my message that I wanted to discuss this and was going to talk about it during the Eng. meeting. 
It's fine only considering the real_data but because we drop the NaN I was thinking that we lose some informations.

npatki:
I think with the above expected algorithm, we wouldn't lose any synthetic data values. Those that go above the real min/max would still be assigned to a bin. We can discuss more soon.

R-Palazzo:
> I think with the above expected algorithm, we wouldn't lose any synthetic data values. Those that go above the real min/max would still be assigned to a bin. We can discuss more soon.

Yes I agree, I like this option. I'm coding it ;)

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,410 @@
...
+        list_dtypes = self._sdtype_to_shape.keys()
+        for column_names in itertools.combinations(list_col_names, r=2):
+            column_name_1 = column_names[0]
+            column_name_2 = column_names[1]
+
+            sdtype_col_1 = metadata['columns'][column_name_1]['sdtype']
+            sdtype_col_2 = metadata['columns'][column_name_2]['sdtype']
+
+            valid_sdtypes = sdtype_col_1 in list_dtypes and sdtype_col_2 in list_dtypes
+            if valid_sdtypes:
```
amontanez24, line 253: to save space with all the indentation, maybe we do 
```python
if not valid_sdtypes:
    continue
```

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,410 @@
...
+            real_data (pandas.DataFrame):
+                The real data
+            synthetic_data (pandas.DataFrame):
+                The synthetic data
+            metadata (dict):
+                The metadata of the table
+            progress_bar:
+                The progress bar to use. Defaults to tqdm.
+        """
+        processed_real_data, processed_synthetic_data = self._get_processed_data(
```
amontanez24, line 232: I wonder if it would be simpler to do the following steps:
1. Process the data (get rid of datetimes)
2. Get the discretized version of the processed data as a separate dataframe (that way you don't need to make unique names)
3. In the loop below, figure out if either column is discrete
4. Based on that, either pass the regular or discretized data to the metric

tests/integration/reports/single_table/_properties/test_column_pair_trends.py
```diff
@@ -0,0 +1,103 @@
...
+        Test the method with the different possible sdtypes as well as with one primary key.
+        """
+        # Setup
+        column_names = [
+            'student_id', 'degree_type', 'start_date', 'second_perc', 'work_experience'
+        ]
+        real_data = pd.read_csv('sdmetrics/demos/single_table/real.csv')[column_names]
+        synthetic_data = pd.read_csv('sdmetrics/demos/single_table/synthetic.csv')[column_names]
+        with open('sdmetrics/demos/single_table/metadata.json', 'r') as f:
+            metadata = json.load(f)
```
amontanez24, line 24: you can probably just use `get_demo`

tests/unit/reports/single_table/_properties/test_column_pair_trends.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # Run
+        cpt_property.get_visualization()
+
+        # Assert
+        assert mock__get_correlation_matrix.call_count == 3
+        mock_make_subplots.assert_called()
+        assert cpt_property._get_heatmap.call_count == 3
+        assert fig_mock.add_trace.call_count == 3
+        cpt_property._update_layout.assert_called()
```
amontanez24, line 311: we should either test `_update_layout` or just test the output here

tests/unit/reports/single_table/_properties/test_column_pair_trends.py
```diff
@@ -0,0 +1,311 @@
...
+            'col1': [1, 2, 3],
+            'col2': [False, True, True],
+            'col3': ['a', 'b', 'c'],
+            'col4': pd.to_datetime(['2020-01-01', '2020-01-02', '2020-01-03'])
+        })
+        synthetic_data = pd.DataFrame({
+            'col1': [4, 5, 6],
+            'col2': [False, True, True],
+            'col3': ['a', 'b', 'c'],
+            'col4': pd.to_datetime(['2020-01-04', '2020-01-05', '2020-01-06'])
```
amontanez24, line 23: should we test with nans?

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,410 @@
...
+        heatmap_df = pd.DataFrame(index=names, columns=names)
+
+        for idx_1, column_name_1 in enumerate(names):
+            for column_name_2 in names[idx_1:]:
+                if column_name_1 == column_name_2:
+                    heatmap_df.loc[column_name_1, column_name_2] = 1
+                    continue
+
+                col_1_cond = (table['Column 1'] == column_name_1)
+                col_2_cond = (table['Column 2'] == column_name_2)
```
amontanez24, line 314: Can we call these col_1_loc and col_2_loc or something like that instead?

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,410 @@
...
+                if table.loc[col_1_cond & col_2_cond].empty:
+                    col_1_cond = (table['Column 1'] == column_name_2)
+                    col_2_cond = (table['Column 2'] == column_name_1)
```
amontanez24, line 317: I'm confused, if one is empty shouldn't the reverse also be?

amontanez24 (commented):
This looks like it was a lot of work! Great job. I think there are a few things we can do to simplify or reduce some calls/conditions. Let's also try to make sure each condition is tested

tests/unit/reports/single_table/_properties/test_column_pair_trends.py
```diff
@@ -0,0 +1,410 @@
...
+        cpt._get_metric('datetime', 'numerical').__name__ == 'CorrelationSimilarity'
+        cpt._get_metric('datetime', 'categorical').__name__ == 'ContingencySimilarity'
+        cpt._get_metric('datetime', 'boolean').__name__ == 'ContingencySimilarity'
+        cpt._get_metric('numerical', 'categorical').__name__ == 'ContingencySimilarity'
+        cpt._get_metric('numerical', 'boolean').__name__ == 'ContingencySimilarity'
+        cpt._get_metric('categorical', 'boolean').__name__ == 'ContingencySimilarity'
+        cpt._get_metric('categorical', 'categorical').__name__ == 'ContingencySimilarity'
+        cpt._get_metric('boolean', 'boolean').__name__ == 'ContingencySimilarity'
+
+    def get_columns_data(self):
```
amontanez24, line 162: test names should start with `test_`

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,441 @@
...
+                The sdtype of the first column
+            sdtype_col_2 (str):
+                The sdtype of the second column
+        """
+        if sdtype_col_1 == 'datetime' or sdtype_col_2 == 'datetime':
+            return True
+        elif self._sdtype_to_shape[sdtype_col_1] != self._sdtype_to_shape[sdtype_col_2]:
+            return True
+        else:
+            return False
```
amontanez24, line 198: Instead of doing all this, I propose just storing two lists:
1. Of column names that failed during datetime
2. Of column names that failed during discretizing
Then when we try to run the metric, we can just see if either column is in the list

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,441 @@
...
+                columns_real = self._get_columns_data(
+                    column_name_1, column_name_2, processed_real_data, metadata
+                )
+                columns_synthetic = self._get_columns_data(
+                    column_name_1, column_name_2, processed_synthetic_data, metadata
+                )
+
+                score_breakdown = metric.compute_breakdown(
+                    real_data=columns_real, synthetic_data=columns_synthetic
+                )
```
amontanez24, line 286: I think the logic in these three methods can be done in one method that just takes in the data, discretized data, column names and metadata. Then it can just check which metric to run and pick the columns from either the original or discretized data

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,441 @@
...
+        )
+
+        for column_name in metadata['columns']:
+            column_meta = metadata['columns'][column_name]
+            column_sdtype = column_meta['sdtype']
+            if column_sdtype in ['numerical', 'datetime']:
+                name_discrete = create_unique_name(column_name + '_discrete', metadata['columns'])
+                processed_data[name_discrete], bin_edges = self._discretize_column(
+                    column_name, processed_data[column_name]
+                )
```
amontanez24, line 130: I think it might be simpler if we returned the discretized data in its own dataframe and maybe with its own function. This way we wouldn't need the `create_unique_name` method at all and we could just use the column names to easily grab the data from the desired dataframe

amontanez24 (changes requested):

sdmetrics/reports/single_table/_properties/column_pair_trends.py
R-Palazzo: Yes I was trying to avoid this for memory saving (not saving twice all the category/boolean columns) but if it's not a concern here, I can return the table (original and discretized one).

sdmetrics/reports/single_table/_properties/column_pair_trends.py
R-Palazzo: Updated in [ebe9e54](https://github.com/sdv-dev/SDMetrics/pull/368/commits/ebe9e54c38ed06f78a6c4ca1a0bf16eb524501e7)

sdmetrics/reports/single_table/_properties/column_pair_trends.py
R-Palazzo: Yes I agree, done in [ebe9e54](https://github.com/sdv-dev/SDMetrics/pull/368/commits/ebe9e54c38ed06f78a6c4ca1a0bf16eb524501e7)

sdmetrics/reports/single_table/_properties/column_pair_trends.py
amontanez24: Why would we need to store all the categorical and boolean columns twice? Can't we just take the columns that you're appending to the table and put only those in a separate dataframe?

frances-h (approved):
Looks good!

tests/integration/reports/single_table/_properties/test_column_pair_trends.py
```diff
@@ -0,0 +1,130 @@
...
+            'Synthetic Correlation': [
+                0.07645134784253645, np.nan, np.nan, np.nan, np.nan, np.nan
+            ]
+        }
+        expected_details = pd.DataFrame(expected_details_dict)
+        pd.testing.assert_frame_equal(column_shape_property._details, expected_details)
+        assert score == 0.754
+
+    def test_get_score_warnings(self, recwarn):
+        """Test the ``get_score`` method when the metrics are raising erros for some columns."""
```
amontanez24, line 60: typo: erros

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,455 @@
...
+                The synthetic data
+            metadata (dict):
+                The metadata of the table
+        """
+        columns_real = self._get_columns_data(
+            col_name_1, col_name_2, real_data, discrete_real, metadata
+        )
+        columns_synthetic = self._get_columns_data(
+            col_name_1, col_name_2, synthetic_data, discrete_synthetic, metadata
+        )
```
amontanez24, line 235: I think you can do the logic for these methods and getting the metric itself all in here since they seem to share similar conditional statements

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,455 @@
...
+        for idx_1, column_name_1 in enumerate(names):
+            for column_name_2 in names[idx_1:]:
+                if column_name_1 == column_name_2:
+                    heatmap_df.loc[column_name_1, column_name_2] = 1
+                    continue
+
+                col_1_loc = (table['Column 1'] == column_name_1)
+                col_2_loc = (table['Column 2'] == column_name_2)
+                if table.loc[col_1_loc & col_2_loc].empty:
+                    col_1_loc = (table['Column 1'] == column_name_2)
```
amontanez24, line 361: Can you leave a small comment here about why we check both combinations?

amontanez24 (approved):
A couple small comments but after that I think this is ready 🍾  Great job!

sdmetrics/reports/single_table/_properties/column_pair_trends.py
```diff
@@ -281,23 +237,22 @@ def _generate_details(self, real_data, synthetic_data, metadata, progress_bar):
 
                 continue
 
-            metric = self._get_metric(sdtype_col_1, sdtype_col_2)
+            columns_real, metric = self._get_columns_data_and_metric(
+                column_name_1, column_name_2, processed_real_data, discrete_real, metadata
+            )
+            columns_synthetic, _ = self._get_columns_data_and_metric(
+                column_name_1, column_name_2, processed_synthetic_data,
+                discrete_synthetic, metadata
```
amontanez24, line 245: Why not get both the synthetic and real data at once?

amontanez24 (approved):
This looks a lot better! I just left one comment but other than that I think this is good!

tests/integration/reports/single_table/_properties/test_column_pair_trends.py
amontanez24: This typo is still here @R-Palazzo

sdmetrics/reports/single_table/_properties/column_pair_trends.py
R-Palazzo: Yes done in [582456b](https://github.com/sdv-dev/SDMetrics/pull/368/commits/582456bbfd5ac86cd5f861c4a420900ea6fa27a0)

sdmetrics/reports/single_table/_properties/column_pair_trends.py
R-Palazzo: Yes I let a comment lint 324-325

PR #369, opened 2023-06-20
Create Parent - Child relationship property class

### LINKED ISSUE #359: Create Parent Child Relationships property class
### Problem Description

Once #357 is done, we can migrate the code for computing the parent child relationship scores to a property class.

#### Attributes

- `metrics`: [CardinalityShapeSimilarity]
- `_properties`: None

#### Abstract methods

- `get_score(real_data, synthetic_data, metadata)` - Average the scores calculated when running the CardinalityShapeSimilarity metric.
- `get_visualization(table_name)` - Just a wrapper around the single table `get_visualization` method.

### Additional context
- Unlike the other properties, this one has no single table equivalent. The metric itself is designed to handle multi table data, so this class will not have to use any single table properties.

pvk-developer:
Resolves #359

### DIFF
--- a/sdmetrics/reports/multi_table/_properties/relationship.py
+++ b/sdmetrics/reports/multi_table/_properties/relationship.py
@@ -1,0 +1,61 @@
+from sdmetrics.multi_table.statistical import CardinalityShapeSimilarity
+from sdmetrics.reports.multi_table._properties.base import BaseMultiTableProperty
+from sdmetrics.reports.multi_table.plot_utils import get_table_relationships_plot
+
+
+class CardinalityShapeSimilarityProperty(BaseMultiTableProperty):
+    """``CardinalityShapeSimilarityProperty`` class.
+
+    Property that uses ``sdmetrics.multi_table.statistical.CardinalityShapeSimilarity`` metric
+    in order to compute and plot the scores of cardinality shape similarity in the given tables.
+    """
+
+    def get_score(self, real_data, synthetic_data, metadata, progress_bar):
+        """Get the average score of cardinality shape similarity in the given tables.
+
+        Args:
+            real_data (pandas.DataFrame):
+                The real data.
+            synthetic_data (pandas.DataFrame):
+                The synthetic data.
+            metadata (dict):
+                The metadata, which contains each column's data type as well as relationships.
+            progress_bar (tqdm.tqdm):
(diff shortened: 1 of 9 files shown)


frances-h (approved):

sdmetrics/reports/multi_table/_properties/relationship.py
```diff
@@ -0,0 +1,76 @@
+from sdmetrics.multi_table.statistical import CardinalityShapeSimilarity
+from sdmetrics.reports.multi_table._properties.base import BaseMultiTableProperty
+from sdmetrics.reports.multi_table.plot_utils import get_table_relationships_plot
+
+
+class CardinalityShapeSimilarityProperty(BaseMultiTableProperty):
```
amontanez24, line 6: We can just call it `Cardinality`. Can we also rename the file to match?

tests/unit/reports/multi_table/_properties/test_relationship.py
```diff
@@ -0,0 +1,89 @@
...
+            ('users', 'sessions'): 1.,
+            ('users', 'transactions'): 0.5
+        }
+
+        # Run
+        result = CardinalityShapeSimilarityProperty.get_visualization(instance, 'sessions')
+
+        # Assert
+        assert result == mock_get_table_relationships_plot.return_value
+        expected_score_breakdwons = {
```
amontanez24, line 84: typo: breakdwons

sdmetrics/reports/multi_table/_properties/relationship.py
```diff
@@ -0,0 +1,76 @@
...
+        for relation in metadata.get('relationships', []):
+            relationships_metadata = {'relationships': [relation]}
+            try:
+                self._metric_results = CardinalityShapeSimilarity.compute_breakdown(
+                    real_data,
+                    synthetic_data,
+                    relationships_metadata
+                )
+            except Exception:
+                pass
```
amontanez24, line 42: I think in the other metrics we are appending the error message to a new column with the header `Error`. We should do the same. That is the old behavior too I believe

amontanez24 (changes requested):

amontanez24 (approved):
LGTM! Thanks for the changes :shipit:

PR #372, opened 2023-06-21
Rename `'Parent Child Relationships'` property to `'Cardinality'`

### LINKED ISSUE #360: In Multi Table Quality Report: Rename `"Table Relationships"` property to `"Cardinality"`
### Problem Description
The [Table Relationships](https://docs.sdv.dev/sdmetrics/reports/quality-report/whats-included#table-relationships) property computes the [CardinalityShapeSimilarity](https://docs.sdv.dev/sdmetrics/metrics/metrics-glossary/cardinalityshapesimilarity) for score for each relationship and average them.

The name "Table Relationships" is really broad. It would be more useful if we renamed it to "Cardinality" to better reflect what is being captured.

### Expected behavior
Change the property name anywhere that it is being used.

- When printing out the results of `report.generate()`
- The results in `get_properties()`
- The property name input in `get_details()`
- The property name input in `get_visualization()`

### Additional context
SDMetrics has not yet had a 1.0 release so we do not officially have to be backwards compatible. But it may be good to do so for the `get_details` and `get_visualization` commands.

If a user puts in `property_name='Table Relationships'`, use the value `'Cardinality'` instead and raise the following warning:
```
FutureWarning: The 'Table Relationships' property name is no longer recognized. Please update to 'Cardinality' instead.
```

frances-h:
Resolve #360

### DIFF
--- a/sdmetrics/reports/multi_table/quality_report.py
+++ b/sdmetrics/reports/multi_table/quality_report.py
@@ -31,7 +31,7 @@ class QualityReport():
     METRICS = {
         'Column Shapes': [KSComplement, TVComplement],
         'Column Pair Trends': [CorrelationSimilarity, ContingencySimilarity],
-        'Parent Child Relationships': [CardinalityShapeSimilarity],
+        'Cardinality': [CardinalityShapeSimilarity],
     }
 
     def __init__(self):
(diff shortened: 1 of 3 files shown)


amontanez24 (approved):
This LGTM!

PR #374, opened 2023-06-23
Create multi table column shapes and column pair trends properties

### LINKED ISSUE #358: Create multi table column shapes and column pair trends properties
### Problem Description

Once #357 is complete, we can implement the actual multi table property classes.

### Expected behavior

Create both a `ColumnPairTrends` and `ColumnShapes` property in the `reports/multi-table` folder. Each property should have its own file.

#### Attributes

- `metrics`: The metrics for each should match that of the `SingleTableProperties`. See #355 and #356 
- `_properties`: A dict mapping the table names to their single table properties.

#### Abstract methods

- `get_score(real_data, synthetic_data, metadata)` - Returns a float that is the average score of all the individual metric scores computed. We can use the `_details` attribute on each single table property to get all the scores for averaging.
- `get_visualization(table_name)` - Just a wrapper around the single table `get_visualization` method.

R-Palazzo:
Resolve #358.

### DIFF
--- a/sdmetrics/reports/multi_table/_properties/column_pair_trends.py
+++ b/sdmetrics/reports/multi_table/_properties/column_pair_trends.py
@@ -1,0 +1,9 @@
+from sdmetrics.reports.single_table._properties import ColumnPairTrends as SingleTableColumnPairTrends
+from sdmetrics.reports.multi_table._properties import BaseMultiTableProperty
+
+
+class ColumnPairTrends(BaseMultiTableProperty):
+
+    def __init__(self):
+        super().__init__()
+        self._single_table_property = SingleTableColumnPairTrends
--- a/sdmetrics/reports/multi_table/_properties/column_pair_trends.py
+++ b/sdmetrics/reports/multi_table/_properties/column_pair_trends.py
@@ -1,8 +1,16 @@
-from sdmetrics.reports.single_table._properties import ColumnPairTrends as SingleTableColumnPairTrends
+"""Column pair trends property for multi-table."""
+from sdmetrics.reports.single_table._properties import (
+    ColumnPairTrends as SingleTableColumnPairTrends)
 from sdmetrics.reports.multi_table._properties import BaseMultiTableProperty
 
 
 class ColumnPairTrends(BaseMultiTableProperty):
+    """Column pair trends property for multi-table.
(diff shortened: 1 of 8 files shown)


amontanez24 (approved):
This LGTM! Great job!

sdmetrics/reports/multi_table/_properties/column_pair_trends.py
```diff
@@ -0,0 +1,17 @@
...
+    """Column pair trends property for multi-table.
+
+    This property evaluates the matching in trends between pairs of real
+    and synthetic data columns. Each pair's correlation is calculated and
+    the final score represents the average of these measures across all column pairs
+    """
+
+    def __init__(self):
+        super().__init__()
+        self._single_table_property = SingleTableColumnPairTrends
```
pvk-developer, line 17: I think that this can be removed and have just `_single_table_property = SingleTableColumnPairTrends` as  attribute of the class and this will still work, no need for the `__init__`

sdmetrics/reports/multi_table/_properties/column_shapes.py
```diff
@@ -0,0 +1,18 @@
...
+    This property assesses the shape similarity between the real and synthetic data.
+    A metric score is computed column-wise and the final score is the average over all columns.
+    The KSComplement metric is used for numerical and datetime columns while the TVComplement
+    is used for categorical and boolean columns.
+    The other column types are ignored by this property.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self._single_table_property = SingleTableColumnShapes
```
pvk-developer, line 18: same as above

pvk-developer (changes requested):
Just change `_single_table_property` to `attribute` rather than in the `instance` and should be good to go.

pvk-developer (approved):

sdmetrics/reports/multi_table/_properties/column_pair_trends.py
R-Palazzo: Yes it would be easier this way, done in [61510e9](https://github.com/sdv-dev/SDMetrics/pull/374/commits/61510e99ff095d8a955a9ab6c1942cb42002d044)

PR #382, opened 2023-07-05
Drop support for Python 3.7

### LINKED ISSUE #380: Drop support for Python 3.7
### Problem Description

Python 3.7 reached its EOL June 27, 2023. As Github drops support some of our builds have started to fail. We should just drop support for 3.7 as a result.

### Expected behavior

- Drop support for 3.7 in setup.py
- Remove all 3.7 builds

amontanez24:
resolves #380


pvk-developer (approved):
Missing `tox.ini` https://github.com/sdv-dev/SDMetrics/blob/06021c4e869deab83ff41f0f1f6db427b86d0f62/tox.ini#L2

After that 🚢

PR #383, opened 2023-07-10
Use property classes in multi table `QualityReport`

### LINKED ISSUE #362: More accurate progress bar for multi table Quality Report
### Problem Description
The current progress bar always increments in 5 steps (no matter how many columns I have) and doesn't give me much insight into the progress. It also pauses after getting to 100% and remains that way for a while (many minutes).

### Expected behavior
The progress indicators should more accurately reflect the computation that is happening. There should be 1 progress bar that's printed _per property_ that indicates the overall progress.

- **Column Shapes**: If there are _n_ total columns in the dataset (sum of all tables), then the progress bar should increment _n_ times.
    - Table A has _a_ columns, then there are _a_ comparisons
    - Table B has _b_ columns, then there are _b_ comparisons
    - ...
    - Total comparisons: _n = a + b + c + ..._
- **Column Pair Trends**: We should be able to compute the total comparisons ahead of time for each table and then sum them.
    - Table A has _a_ columns, then there are _a(a-1)/2_ comparisons
    - Table B has _b_ columns, then there are _b(b-1)/2_ comparisons
    - ... 
    - Total comparisons: _[a(a-1)/2] + [b(b-1)/2] + [c(c-1)/2] + ..._
- **Cardinality**: If there are _r_ relationships, then the progress bar should increment _r_ times.

### Example
```python
```
```
Generating report ...

(1/3) Evaluating Column Shapes: 100%|██████████| 250/250 [00:00<00:00,  7.09it/s]
(2/3) Evaluating Column Pair Trends: 100%|██████████| 1200/1200 [00:00<00:00,  7.09it/s]
(3/3) Evaluating Cardinality: 100%|██████████| 5/5 [00:00<00:00,  7.09it/s]

Overall Quality Score: 82.84%

Properties:
- Column Shapes: 82.78%
- Column Pair Trends: 82.9%
- Cardinality: 77%
```

### LINKED ISSUE #371: Use property classes in multi table QualityReport
### Problem Description

Once #358 and #359 are complete, we should use the property classes in the `QualityReport` itself.

### Expected behavior

- The `get_details` method should just return the details from the desired property class.
- The `get_visualization` should just be a wrapper for the property class' `get_visualization`
- The `generate` method should loop over the properties and use a class for each one to get the property scores and then average in the end.

### Additional context

- The `get_raw_result` method can be removed
- All other functionality should remain the same

fealho:
Resolve #371, resolve #362


R-Palazzo (commented):
Looking good! The code is much lighter now haha.
I let few comments.

sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -71,68 +37,31 @@ def generate(self, real_data, synthetic_data, metadata, verbose=True):
                 The metadata, which contains each column's data type as well as relationships.
             verbose (bool):
                 Whether or not to print report summary and progress.
+                NOTE: todo
```
amontanez24, line 40: If this is going to be in another issue can we say what issue here? Otherwise let's just implement it

sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -226,124 +138,24 @@ def get_details(self, property_name, table_name=None):
...
-        return details
+        property_instance = self._properties_instances[property_name]
+        if property_name != 'Cardinality':
+            if table_name:
+                return property_instance._properties[table_name]._details.copy()
 
-    def get_raw_result(self, metric_name):
-        """Return the raw result of the given metric name.
+            details = {}
+            for table_name, properties in property_instance._properties.items():
```
amontanez24, line 149: minor: I think it should be table_name, property since you're dealing with a single property at this point

tests/integration/reports/multi_table/test_multi_table_quality_report.py
```diff
@@ -77,13 +78,75 @@ def load_test_data():
 
 def test_multi_table_quality_report():
     """Test the multi table quality report."""
```
amontanez24, line 80: this massive test could definitely use a descriptive docstring

tests/unit/reports/multi_table/test_multi_table_quality_report.py
```diff
@@ -22,610 +22,257 @@ def test___init__(self):
...
+    def test_get_properties(self):
+        """Test the ``get_properties`` method."""
+        # Setup
+        report = QualityReport()
+        mock_properties_scores = {
+            'Column Shapes': 0.1,
+            'Column Pair Trends': 0.2,
+            'Cardinality': 0.3,
         }
+        report._properties_scores = mock_properties_scores
```
amontanez24, line 114: you can just do this in one assignment haha

tests/unit/reports/multi_table/test_multi_table_quality_report.py
```diff
@@ -22,610 +22,257 @@ def test___init__(self):
...
+        report = QualityReport()
+        instance = Mock()
+        instance._details = {'details'}
+        report._properties_instances = {'Cardinality': instance}
+        report._is_generated = True
 
-        Setup:
-        - Mock the ``_overall_quality_score`` attribute.
+        # Run
+        details = report.get_details('Cardinality')
```
amontanez24, line 216: we should add a test for when the property is cardinality but they give a table name to make sure the filtering works

amontanez24 (commented):
Looking good! Mostly minor comments

amontanez24 (approved):
I left one comment about the unit test mock but after that this looks good! Thanks for addressing the comments :shipit:

R-Palazzo (approved):
LGTM!

PR #385, opened 2023-07-16
More accurate progress bar for multi table Quality Report

### LINKED ISSUE #362: More accurate progress bar for multi table Quality Report

fealho:
Resolve #362.

### DIFF
--- a/sdmetrics/reports/multi_table/quality_report.py
+++ b/sdmetrics/reports/multi_table/quality_report.py
@@ -1,11 +1,13 @@
 """Multi table quality report."""
 
 import pickle
+import sys
 import warnings
 
 import numpy as np
 import pandas as pd
 import pkg_resources
+import tqdm
 
 from sdmetrics.reports.multi_table._properties import Cardinality, ColumnPairTrends, ColumnShapes
 from sdmetrics.reports.utils import validate_multi_table_inputs
@@ -25,6 +27,26 @@ class QualityReport():
         self._properties_scores = {}
         self._is_generated = False
         self._package_version = None
+        self._property_errors = {}
+
+    def _print_results(self, out=sys.stdout):
+        """Print the quality report results."""
+        if pd.isna(self._overall_quality_score) & any(self._property_errors.values()):
+            out.write('\nOverall Quality Score: Error computing report.\n\n')
+        else:
+            out.write(
+                f'\nOverall Quality Score: {round(self._overall_quality_score * 100, 2)}%\n\n')
+
+        if len(self._properties_instances) > 0:
+            out.write('Properties:\n')
+
+        for property_name, score in self._properties_scores.items():
+            if not pd.isna(score):
+                out.write(f'{property_name}: {round(score * 100, 2)}%\n')
+            elif property_name in self._property_errors:
+                out.write(f'{property_name}: Error computing property.\n')
+            else:
+                out.write(f'{property_name}: NaN\n')
 
     def generate(self, real_data, synthetic_data, metadata, verbose=True):
         """Generate report.
@@ -37,8 +59,7 @@ class QualityReport():
             metadata (dict):
                 The metadata, which contains each column's data type as well as relationships.
             verbose (bool):
-                Whether or not to print report summary and progress.
-                NOTE: todo in GitHub Issue #362.
+                Whether or not to print the report summary and progress.
         """
         validate_multi_table_inputs(real_data, synthetic_data, metadata)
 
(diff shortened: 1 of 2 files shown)


sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -25,6 +27,26 @@ def __init__(self):
         self._properties_scores = {}
         self._is_generated = False
         self._package_version = None
+        self._property_errors = {}
+
+    def _print_results(self, out=sys.stdout):
+        """Print the quality report results."""
+        if pd.isna(self._overall_quality_score) & any(self._property_errors.values()):
+            out.write('\nOverall Quality Score: Error computing report.\n\n')
```
fealho, line 35: Run this line whenever `_overall_quality_score` is `nan`, so we don't need `_property_errors`.

sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -25,6 +27,26 @@ def __init__(self):
...
+        else:
+            out.write(
+                f'\nOverall Quality Score: {round(self._overall_quality_score * 100, 2)}%\n\n')
+
+        if len(self._properties_instances) > 0:
```
fealho, line 40: Remove this check.

sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -25,6 +27,26 @@ def __init__(self):
...
+            out.write('Properties:\n')
+
+        for property_name, score in self._properties_scores.items():
+            if not pd.isna(score):
+                out.write(f'{property_name}: {round(score * 100, 2)}%\n')
+            elif property_name in self._property_errors:
```
fealho, line 46: Remove this, so we always just return it as `nan` (or remove the check below, so we always output an error when the score is `nan`). This let's me remove the `_property_errors` attribute and simplify the `generate` code a bit.

fealho (commented):
@amontanez24 What should we do with the following messages? I don't think they are necessary anymore, and removing them would make the code a little lighter.

sdmetrics/reports/multi_table/quality_report.py
amontanez24: Yeah I think we can remove this since the method is only called in generate

sdmetrics/reports/multi_table/quality_report.py
amontanez24: I think you can get nan without it being an error

sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -50,15 +70,45 @@ def generate(self, real_data, synthetic_data, metadata, verbose=True):
...
+                elif property_name == 'Column Pair Trends':
+                    # for each table, the number of combinations of pairs of columns is
+                    # n * (n - 1) / 2, where n is the number of columns in the table
+                    num_iterations = sum([(n_cols * (n_cols - 1)) // 2 for n_cols in num_columns])
+                elif property_name == 'Cardinality':
+                    num_iterations = len(metadata['relationships'])
+
+                progress_bar = tqdm.tqdm(total=num_iterations, file=sys.stdout)
+                progress_bar.set_description(
+                    f'({index + 1}/{num_properties}) Evaluating {property_name}: ')
```
amontanez24, line 93: are there unit tests for all this? That make sure the right description is set

sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -50,15 +70,45 @@ def generate(self, real_data, synthetic_data, metadata, verbose=True):
...
+
+            try:
+                self._properties_scores[property_name] = property_instance.get_score(
+                    real_data, synthetic_data, metadata, progress_bar)
+            except BaseException:
+                self._properties_scores[property_name] = np.nan
+                self._property_errors[property_name] = True
```
amontanez24, line 100: we should also have unit tests for when get_score raises an exception

amontanez24 (commented):
Looks good! Just unsure if we tested everything

R-Palazzo (approved):
LGTM!

tests/unit/reports/multi_table/test_multi_table_quality_report.py
```diff
@@ -85,6 +179,46 @@ def test_generate(self, mock_cardinality_score, mock_column_shapes_score,
...
+        """Test the ``generate`` method when `get_score` for each property fails."""
+        # Setup
+        real_data, synth_data, metadata = self.get_data()
+        report = QualityReport()
+        mock_cardinality_score.side_effect = ValueError
+        mock_column_shapes_score.side_effect = ValueError
+        mock_column_pair_trends_score.side_effect = ValueError
+
+        # Run
+        report.generate(real_data, synth_data, metadata)
```
amontanez24, line 209: can we either 
a. add a test where verbose is true that checks that `_print_results` gets called
or
b. add a test where verbose is true and just move the asserts from test__print_results into that test

amontanez24 (commented):
One more test request but after that :shipit:

PR #387, opened 2023-07-18
Add support for Python 3.11

### LINKED ISSUE #353: Add support for Python 3.11
### Problem Description
SDMetrics should work on Python 3.11

### Background
The SDMetrics library currently runs on Python versions 3.7-3.10. A while ago, we had made the decision to exclude 3.11 because the most recent versions of torch were incompatible with it.

However, torch is now fully upgraded to 2.0 so this should no longer be an issue.

amontanez24:
resolves #353


fealho (approved):
Change it in tox.ini as well, looks good otherwise 👍

frances-h (approved):
Looks good! 👍

PR #396, opened 2023-07-20
Add add-on detection for premium metrics

### LINKED ISSUE #388: Add add-on detection for premium metrics
### Problem Description

_As an enterprise customer, it would be convenient to be able to access enterprise SDMetrics features in the same way as public ones._

### Expected behavior

- We should be able to import enterprise metrics in the same way that we import public ones.

```python
from sdmetrics.single_column import EnterpriseSingleColumnMetrics
```

### Additional context

- We should use a style similar to what is done [here](https://github.com/sdv-dev/RDT/blob/06a3bddfb7ab664723b4bb4b90687c9fa40a2b4c/rdt/__init__.py#L143)

amontanez24:
resolves #388


sdmetrics/__init__.py
```diff
@@ -6,11 +6,16 @@
 __email__ = 'dailabmit@gmail.com'
 __version__ = '0.10.2.dev0'
 
+import sys
+import warnings as python_warnings
```
amontanez24, line 10: I had to do this because in the test when importing sdmetrics, warnings was being overridden by [sdmetrics.warnings](https://github.com/sdv-dev/SDMetrics/blob/master/sdmetrics/warnings.py)

frances-h (approved):

fealho (approved):

PR #398, opened 2023-07-21
Create single table synthesis property

### LINKED ISSUE #390: Create single table synthesis property
### Problem Description

Now that we have property classes, we want to move the metrics used in the `DiagnosticReport` into their respective properties.

This is the spec for the `Synthesis` property

#### Attributes

- `_details` - A dataframe containing the following columns
    - 'Metric'
    - 'Score'
    - 'Num Matched Rows'
    - 'Num New Rows'

The metrics used in this class are: [NewRowSynthesis]

#### Methods

- `_generate_details(self, real_data, synthetic_data, metadata, progress_bar)` - Generate the `_details` dataframe that will be used by the base class to get the score.
Pseudocode:
```python
score = NewRowSynthesis.compute_breakdown(real_data, synthetic_data, synthetic_sample_size: 10000)
```
- `get_visualization()` - Returns a plotly.graph_objects._figure.Figure object for the specified table. Use code similar to what's in [get_synthesis_plot](https://github.com/sdv-dev/SDMetrics/blob/master/sdmetrics/reports/single_table/plot_utils.py#L390).

### Expected behavior

```python
>>> prop = Synthesis()
>>> prop.get_score()
>>> 0.89
>>> prop.get_visualization()
```

### Additional context

- we should catch any exceptions the same way we do [here](https://github.com/sdv-dev/SDMetrics/blob/master/sdmetrics/reports/single_table/_properties/column_shapes.py)

R-Palazzo:
Resolve #390

### DIFF
--- a/sdmetrics/reports/single_table/_properties/synthesis.py
+++ b/sdmetrics/reports/single_table/_properties/synthesis.py
@@ -34,10 +34,8 @@ class Synthesis(BaseSingleTableProperty):
         """
         name = self.metric.__name__
         error_message = np.nan
-        if len(synthetic_data) > 10000:
-            sample_size = 10000
-        else:
-            sample_size = len(synthetic_data)
+
+        sample_size = len(synthetic_data) if len(synthetic_data) < 10000 else 10000
 
         try:
             score_breakdown = self.metric.compute_breakdown(
--- a/sdmetrics/reports/single_table/_properties/synthesis.py
+++ b/sdmetrics/reports/single_table/_properties/synthesis.py
@@ -7,7 +7,12 @@ from sdmetrics.single_table import NewRowSynthesis
 
 
 class Synthesis(BaseSingleTableProperty):
-    """Synthesis property class for single table."""
+    """Synthesis property class for single table.
+
+    This property assesses the novelty of the syntetic data over the real data.
+    The NewRowSynthesis metric is computed over the real and synthetic table to
+    score the proportion of new rows in the synthetic data.
+    """
 
     metric = NewRowSynthesis
 
--- a/sdmetrics/reports/single_table/_properties/synthesis.py
+++ b/sdmetrics/reports/single_table/_properties/synthesis.py
@@ -17,7 +17,7 @@ class Synthesis(BaseSingleTableProperty):
     metric = NewRowSynthesis
 
     def _generate_details(self, real_data, synthetic_data, metadata, progress_bar=None):
-        """Generate the _details dataframe for the column shapes property.
+        """Generate the _details dataframe for the synthesis property.
 
         Args:
             real_data (pandas.DataFrame):
--- a/sdmetrics/reports/single_table/_properties/synthesis.py
+++ b/sdmetrics/reports/single_table/_properties/synthesis.py
@@ -36,7 +36,6 @@ class Synthesis(BaseSingleTableProperty):
         error_message = np.nan
 
         sample_size = len(synthetic_data) if len(synthetic_data) < 10000 else 10000
-
         try:
             score_breakdown = self.metric.compute_breakdown(
                 real_data, synthetic_data, synthetic_sample_size=sample_size
--- a/sdmetrics/reports/single_table/_properties/synthesis.py
+++ b/sdmetrics/reports/single_table/_properties/synthesis.py
@@ -21,16 +21,16 @@ class Synthesis(BaseSingleTableProperty):
 
         Args:
             real_data (pandas.DataFrame):
-                The real data
+                The real data.
             synthetic_data (pandas.DataFrame):
-                The synthetic data
+                The synthetic data.
             metadata (dict):
-                The metadata of the table
+                The metadata of the table.
             progress_bar (tqdm.tqdm or None):
                 The progress bar to use. Defaults to tqdm.
 
         Returns:
-            pandas.DataFrame
+            pandas.DataFrame.
         """
         name = self.metric.__name__
         error_message = np.nan
@@ -65,7 +65,7 @@ class Synthesis(BaseSingleTableProperty):
             'Error': error_message,
         }, index=[0])
 
-        if result['Error'].isna().all():
+        if pd.isna(result['Error'].iloc[0]):
             result = result.drop('Error', axis=1)
 
         return result
@@ -74,14 +74,12 @@ class Synthesis(BaseSingleTableProperty):
         """Create a plot to show the synthesis property.
 
         Returns:
-            plotly.graph_objects._figure.Figure
+            plotly.graph_objects._figure.Figure.
         """
         labels = ['Exact Matches', 'Novel Rows']
-        values = list(self._details[['Num Matched Rows', 'Num New Rows']].to_numpy()[0])
+        values = list(self._details[['Num Matched Rows', 'Num New Rows']].iloc[0])
 
-        average_score = self._compute_average()
-        if not np.isnan(average_score):
-            average_score = round(average_score, 2)
+        average_score = round(self._compute_average(), 2)
 
         fig = px.pie(
             values=values,
--- a/sdmetrics/reports/single_table/_properties/synthesis.py
+++ b/sdmetrics/reports/single_table/_properties/synthesis.py
@@ -28,7 +28,7 @@ class Synthesis(BaseSingleTableProperty):
             pandas.DataFrame
         """
         name = self.metric.__name__
-        error_message = None
+        error_message = np.nan
         if len(synthetic_data) > 10000:
             sample_size = 10000
         else:
--- a/sdmetrics/reports/single_table/_properties/synthesis.py
+++ b/sdmetrics/reports/single_table/_properties/synthesis.py
@@ -10,7 +10,7 @@ class Synthesis(BaseSingleTableProperty):
     """Synthesis property class for single table.
 
     This property assesses the novelty of the syntetic data over the real data.
-    The NewRowSynthesis metric is computed over the real and synthetic table to
+    The ``NewRowSynthesis`` metric is computed over the real and synthetic table to
     score the proportion of new rows in the synthetic data.
     """
 
--- a/sdmetrics/reports/single_table/_properties/synthesis.py
+++ b/sdmetrics/reports/single_table/_properties/synthesis.py
@@ -1,0 +1,95 @@
+import numpy as np
+import pandas as pd
+import plotly.express as px
+
+from sdmetrics.reports.single_table._properties import BaseSingleTableProperty
+from sdmetrics.single_table import NewRowSynthesis
+
+
+class Synthesis(BaseSingleTableProperty):
+    """Synthesis property class for single table."""
+
+    metric = NewRowSynthesis
+
+    def _generate_details(self, real_data, synthetic_data, metadata, progress_bar=None):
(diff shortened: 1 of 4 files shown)


sdmetrics/reports/single_table/_properties/synthesis.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    metric = NewRowSynthesis
+
+    def _generate_details(self, real_data, synthetic_data, metadata, progress_bar=None):
+        """Generate the _details dataframe for the synthesis property.
+
+        Args:
+            real_data (pandas.DataFrame):
+                The real data
```
fealho, line 24: We usually end args with dots. Same for args below.

sdmetrics/reports/single_table/_properties/synthesis.py
```diff
@@ -0,0 +1,100 @@
...
+                The metadata of the table
+            progress_bar (tqdm.tqdm or None):
+                The progress bar to use. Defaults to tqdm.
+
+        Returns:
+            pandas.DataFrame
+        """
+        name = self.metric.__name__
+        error_message = np.nan
+        if len(synthetic_data) > 10000:
```
fealho, line 37: Why is this here? And should this not log some sort of warning?

It can also be written as one line: `sample_size = len(synthetic_data) if len(synthetic_data) <= 10000 else 10000`

sdmetrics/reports/single_table/_properties/synthesis.py
```diff
@@ -0,0 +1,100 @@
...
+            score = np.nan
+            num_matched_rows = np.nan
+            num_new_rows = np.nan
+            error_message = f'Error: {type(e).__name__} {e}'
+
+        finally:
+            if progress_bar:
+                progress_bar.update()
+
+        result = pd.DataFrame({
```
fealho, line 60: Are any of the objects a list? If not, it's weird to me we are using a DataFrame when a simple dict would suffice.

sdmetrics/reports/single_table/_properties/synthesis.py
```diff
@@ -0,0 +1,100 @@
...
+            'Metric': name,
+            'Score': score,
+            'Num Matched Rows': num_matched_rows,
+            'Num New Rows': num_new_rows,
+            'Error': error_message,
+        }, index=[0])
+
+        if result['Error'].isna().all():
```
fealho, line 68: Can result['Error'] be a list? If not we don't need the `all`.

I also find it more intuitive to add the errors to `result` if there are any, rather than removing them if there were none.

sdmetrics/reports/single_table/_properties/synthesis.py
```diff
@@ -0,0 +1,100 @@
...
+        return result
+
+    def get_visualization(self):
+        """Create a plot to show the synthesis property.
+
+        Returns:
+            plotly.graph_objects._figure.Figure
+        """
+        labels = ['Exact Matches', 'Novel Rows']
+        values = list(self._details[['Num Matched Rows', 'Num New Rows']].to_numpy()[0])
```
fealho, line 80: I don't think you need `to_numpy`, you can select the first row with `iloc[0]`.

sdmetrics/reports/single_table/_properties/synthesis.py
```diff
@@ -0,0 +1,100 @@
...
+
+        average_score = self._compute_average()
+        if not np.isnan(average_score):
```
fealho, line 83: I don't think you need this check, `round` works for nans.

tests/unit/reports/single_table/_properties/test_synthesis.py
```diff
@@ -0,0 +1,94 @@
...
+                'col4': {'sdtype': 'datetime'}
+            }
+        }
+
+        newrowsynthesis_mock.return_value = {
+            'score': 0.25,
+            'num_matched_rows': 3,
+            'num_new_rows': 1,
+        }
+        # Run
```
fealho, line 41: New line

tests/unit/reports/single_table/_properties/test_synthesis.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class TestSynthesis:
+
+    @patch('sdmetrics.reports.single_table._properties.synthesis.'
+           'NewRowSynthesis.compute_breakdown')
+    def test__generate_details(self, newrowsynthesis_mock):
+        """Test the ``_generate_details`` method."""
+        # Setup
+        real_data = pd.DataFrame({
```
fealho, line 15: I don't think you need the real_data/metadata block, since they are not used for anything (just pass empty dataframes or mocks), and synthetic_data can just be a one liner, since it only checks the size of that. Just to make it more obvious that what's actually been tested is the return of the mock.

tests/unit/reports/single_table/_properties/test_synthesis.py
```diff
@@ -0,0 +1,94 @@
...
+        # Run
+        synthesis_property = Synthesis()
+        details = synthesis_property._generate_details(real_data, synthetic_data, metadata)
+
+        # Assert
+        expected_calls = [
+            call(real_data, synthetic_data, synthetic_sample_size=4)
+        ]
+
+        newrowsynthesis_mock.assert_has_calls(expected_calls)
```
fealho, line 50: You also need to test when synthetic data has over 10,000 rows.

fealho (changes requested):
Looking good, just a few suggestions.

sdmetrics/reports/single_table/_properties/synthesis.py
R-Palazzo: Yes, done in [b00d7ca](https://github.com/sdv-dev/SDMetrics/pull/398/commits/b00d7cafd54bf24f3b76c78763bddbd65d5b9408)

sdmetrics/reports/single_table/_properties/synthesis.py
R-Palazzo: @npatki I was also not sure about this.
For the `NewRowSynthesis`, by default, we limit the `sample size` of the synthetic data to 10000 right?
Should we raise a log or a warning if the synthetic data is larger than 10000 rows?

sdmetrics/reports/single_table/_properties/synthesis.py
R-Palazzo: It's to keep the same abstraction as the other properties. `_generate_details` should generate the details table, which is a DataFrame. I agree it's a bit weird because there is only 1 row.

sdmetrics/reports/single_table/_properties/synthesis.py
R-Palazzo: Yes, done in [b00d7ca](https://github.com/sdv-dev/SDMetrics/pull/398/commits/b00d7cafd54bf24f3b76c78763bddbd65d5b9408)

sdmetrics/reports/single_table/_properties/synthesis.py
R-Palazzo: Yes, done in [b00d7ca](https://github.com/sdv-dev/SDMetrics/pull/398/commits/b00d7cafd54bf24f3b76c78763bddbd65d5b9408)

tests/unit/reports/single_table/_properties/test_synthesis.py
R-Palazzo: Yes, done in [b00d7ca](https://github.com/sdv-dev/SDMetrics/pull/398/commits/b00d7cafd54bf24f3b76c78763bddbd65d5b9408)

tests/unit/reports/single_table/_properties/test_synthesis.py
R-Palazzo: Yes, done in [b00d7ca](https://github.com/sdv-dev/SDMetrics/pull/398/commits/b00d7cafd54bf24f3b76c78763bddbd65d5b9408)

tests/unit/reports/single_table/_properties/test_synthesis.py
R-Palazzo: Yes, done in [b00d7ca](https://github.com/sdv-dev/SDMetrics/pull/398/commits/b00d7cafd54bf24f3b76c78763bddbd65d5b9408)

sdmetrics/reports/single_table/_properties/synthesis.py
amontanez24: I don't think we need this here. Just pass it in as 10000 to the metric. If the data has more than 10000 rows it will subsample, otherwise it will use all the data

tests/unit/reports/single_table/_properties/test_synthesis.py
```diff
@@ -0,0 +1,86 @@
...
+import pandas as pd
+
+from sdmetrics.reports.single_table._properties.synthesis import Synthesis
+
+
+class TestSynthesis:
+
+    @patch('sdmetrics.reports.single_table._properties.synthesis.'
+           'NewRowSynthesis.compute_breakdown')
+    def test__generate_details(self, newrowsynthesis_mock):
```
amontanez24, line 12: can we add a unit test for the error case as well

amontanez24 (changes requested):

tests/unit/reports/single_table/_properties/test_synthesis.py
R-Palazzo: Yes I was doing it hahaha. Done in [a310a75](https://github.com/sdv-dev/SDMetrics/pull/398/commits/a310a7541be76a3cc5fac3b80250bbc1b7f8c665)

sdmetrics/reports/single_table/_properties/synthesis.py
R-Palazzo: I did this not to raise this warning (because the user no longer has the possibility to adjust the `sample_size`): https://github.com/sdv-dev/SDMetrics/blob/ddec4f76e3cb866d13b8aded0e319b3e4e96ca3d/sdmetrics/single_table/new_row_synthesis.py#L68

amontanez24: Or you can leave it but make it onle line like @fealho suggested

R-Palazzo: Yes alright, done in [4669397](https://github.com/sdv-dev/SDMetrics/pull/398/commits/4669397521b9df075bc08a7d8a4fbbd880177250)

amontanez24 (approved):
LGTM!

tests/unit/reports/single_table/_properties/test_synthesis.py
```diff
@@ -50,6 +51,38 @@ def test__generate_details(self, newrowsynthesis_mock):
 
         pd.testing.assert_frame_equal(details, expected__details)
 
+    @patch('sdmetrics.reports.single_table._properties.synthesis.'
+           'NewRowSynthesis.compute_breakdown')
+    def test__generate_details_error(self, newrowsynthesis_mock):
+        """Test the ``_generate_details`` method when the metric raises an error."""
+        # Setup
+
```
fealho, line 59: Delete new line

fealho (approved):
Thanks for addressing all those comments!

PR #410, opened 2023-07-27
Don't modify the rounding in the quality report

### LINKED ISSUE #401: Don't modify the rounding in the quality report
_SDMetrics version: Development (master branch)_

The dev version of the quality report shows scores that are rounded differently than before. We should avoid making unnecessary changes like this and stick to what we currently have (in the 0.10.1 release).

We have observed incorrect rounding in 2 places.

### The get_properties() function
This function should return a DataFrame with the _full_ property score. It is the only way a user can programmatically access each property score so we should make sure they can access the detailed float value.

**Observed**: In the dev version, the scores are rounded to 2 decimal digits
image

**Expected**: Don't change what was in 0.10.1. The method should return the full score
image

### Verbosity in the generate() method
As a convenience to the user, we also display the total score and property scores in the `generate()` method if the verbosity is true. In this case, we should avoid making unnecessary changes to the rounding, and stick to what we had before.

**Observed**: In the dev version, all scores (overall and property scores) are rounded only to 1 decimal digit
image

**Expected**: Don't change what was in 0.10.1. The display should be rounded to 2 decimal digits
image

R-Palazzo:
Resolve #401
Added a fix for the multi-table to handle `NaN` in the property score.

### DIFF
--- a/sdmetrics/reports/multi_table/_properties/base.py
+++ b/sdmetrics/reports/multi_table/_properties/base.py
@@ -49,7 +49,7 @@ class BaseMultiTableProperty():
 
         self.is_computed = True
 
-        return round(average_score.mean(), 2)
+        return np.nanmean(average_score)
 
     def get_visualization(self, table_name):
         """Return a visualization for each score in the property.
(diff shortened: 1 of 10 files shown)


amontanez24 (approved):
I think this looks good!

fealho (approved):

PR #411, opened 2023-07-28
Use property classes in single table DiagnosticReport

### LINKED ISSUE #392: Use property classes in single table DiagnosticReport
### Problem Description

Once #389 , #390 and #391  are done, we should use these properties in the [DiagnosticReport](https://github.com/sdv-dev/SDMetrics/blob/master/sdmetrics/reports/single_table/diagnostic_report.py).

### Expected behavior

- The code should loop over the property classes, compute the scores and use that to create the diagnostic report. All user facing functionality should stay the same with the exception of progress bars.
- Similarly to what was done in the [QualityReport](https://github.com/sdv-dev/SDMetrics/blob/f59fdae145cdad5f4216c47f812ad154f1945256/sdmetrics/reports/single_table/quality_report.py#L113), we should base the progress bar off of each property and compute the total length based on what the property loops over.

### Additional context

- The implementation should draw influence from the QualityReport refactoring done in #377 and #378.

R-Palazzo:
Resolve #392
Following our discussion, I created a `BaseReport` class. Now the `QualityReport` and the `DiagnosticReport` inherit from it.
Then I tried to stick to the documentation to define the `DiagnosticReport`. 
Few things I noticed:
- Currently, the `get_result()` method does return something only if `verbose` is `True `(default). Is it the expected behavior or should we always be able to return the dict when one call `get_result()`? @npatki 
- This PRs implies few changes to the documentation (maybe already planed).
-- Change the example usage for `generate()` to show the new `progress bar`
--  `get_properties()` in `DiagnosticReport` returns now a `pandas.DataFrame` (as the `QualityReport`) and not a `dict`
-- In the example of the usage of `get_details()` the third column is now `Score` and not `Diagnostic Score`.
- Finally just to double-check, the `DiagnosticReport` does not have a `get_score` method, is it expected?


npatki:
@R-Palazzo, the `get_result()` method should always return results so I would update this. In general, the `verbose` parameter **only** refers to what is printed out during the most computationally intensive function (generate, fit, etc.). The `verbose` parameter should _not_ affect how any other functions work.

tests/unit/reports/single_table/test_single_table_base_report.py
```diff
@@ -0,0 +1,398 @@
...
+        metadata['columns']['column3'] = {'sdtype': 'numerical'}
+        metadata['columns']['column4'] = {'sdtype': 'numerical'}
+
+        base_report._validate_metadata_matches_data(real_data, synthetic_data, metadata)
+
+    def test__validate_metadata_matches_data_no_mismatch(self):
+        """Test the ``_validate_metadata_matches_data`` method.
+
+        This test checks that the method does not raise an error when there is no column mismatch
+        between the data and the metadata
```
amontanez24, line 62: nitpick: add period at the end

tests/unit/reports/single_table/test_single_table_base_report.py
```diff
@@ -0,0 +1,398 @@
...
+        # Assert
+        mock__validate_metadata_matches_data.assert_called_once_with(
+            real_data, synthetic_data, metadata
+        )
+        mock_validate_categorical_values.assert_called_once_with(
+            real_data, synthetic_data, metadata
+        )
+
+    def test_generate(self):
+        """Test the ``generate`` method."""
```
amontanez24, line 124: maybe we can describe this test a little more

tests/unit/reports/single_table/test_single_table_diagnostic_report.py
```diff
@@ -1,560 +1,101 @@
...
-
-    def test_get_details_synthesis(self):
-        """Test the ``get_details`` method wit synthesis metrics.
+        assert num_iterations_coverage == 3
+        assert num_iterations_boundaries == 3
+        assert num_iterations_synthesis == 1
 
-        Expect that the details of the desired property is returned.
-        """
+    def test_get_results(self):
```
amontanez24, line 47: minor: add docstring

amontanez24 (approved):
This looks great! We can maybe improve the docstrings in some tests but besides that :shipit:

R-Palazzo:
Thanks for the review @amontanez24. I addressed the comments in [e0fdaa7](https://github.com/sdv-dev/SDMetrics/pull/411/commits/e0fdaa7a8ab77614a72085c0c66b12edc9265f5b). I also fixed `get_results()`

sdmetrics/reports/single_table/diagnostic_report.py
```diff
@@ -49,31 +72,29 @@ def _get_num_iterations(self, property_name, metadata):
...
+                The real data.
+            synthetic_data (pandas.DataFrame):
+                The synthetic data.
+            metadata (dict):
+                The metadata, which contains each column's data type.
+            verbose (bool):
+                Whether or not to print report summary and progress.
+        """
+        super().generate(real_data, synthetic_data, metadata, verbose)
+        self._generate_results()
```
amontanez24, line 96: I don't think this will work right because `print_results` gets called before `_generate_results` when verbose is True, so I don't think it will print correctly

sdmetrics/reports/single_table/base_report.py
```diff
@@ -0,0 +1,221 @@
...
+import pickle
+import sys
+import warnings
+
+import numpy as np
+import pandas as pd
+import pkg_resources
+import tqdm
+
+from sdmetrics.reports.utils import _validate_categorical_values
```
pvk-developer, line 11: I'm not sure if there is some sort of `pep` or statement in the [google styleguide](https://google.github.io/styleguide/pyguide.html) but if importing a method from other files I tend to make it public (removing the `_` and adding documentation to it.

sdmetrics/reports/single_table/base_report.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def _print_results(self, out=sys.stdout):
+        raise NotImplementedError
+
+    def _get_num_iterations(self, property_name, metadata):
+        raise NotImplementedError
+
+    def generate(self, real_data, synthetic_data, metadata, verbose=True):
+        """Generate report.
+
```
pvk-developer, line 79: Can we add a short description, eg: Iterate over all the `properties` and compute the scores etc..

sdmetrics/reports/single_table/base_report.py
```diff
@@ -0,0 +1,221 @@
...
+        Returns:
+            pandas.DataFrame
+                The property score.
+        """
+        self._check_report_generated()
+        name, score = [], []
+        for property_name in self._properties:
+            name.append(property_name)
+            score.append(self._properties[property_name]._compute_average())
+        return pd.DataFrame({
```
pvk-developer, line 153: add a breakline before this line

sdmetrics/reports/single_table/base_report.py
```diff
@@ -0,0 +1,221 @@
...
+
+        Returns:
+            pandas.DataFrame
+                The property score.
+        """
+        self._check_report_generated()
+        name, score = [], []
+        for property_name in self._properties:
+            name.append(property_name)
+            score.append(self._properties[property_name]._compute_average())
```
pvk-developer, line 152: you can iterate over `items` and there you will have the property and the property name:
```python
for property_name, property in self._properties.items():
    name.append(property_name)
    score.append(property._compute_average())
```

sdmetrics/reports/single_table/base_report.py
```diff
@@ -0,0 +1,221 @@
...
+        Args:
+            property_name (str):
+                The name of the property to return score details for.
+
+        Returns:
+            plotly.graph_objects._figure.Figure
+                The visualization for the requested property.
+        """
+        self._validate_property_generated(property_name)
+
```
pvk-developer, line 170: can remove this extra blank line

sdmetrics/reports/single_table/base_report.py
```diff
@@ -0,0 +1,221 @@
...
+        Args:
+            property_name (str):
+                The name of the property to return details for.
+
+        Returns:
+            pandas.DataFrame
+        """
+        self._validate_property_generated(property_name)
+
```
pvk-developer, line 184: can remove this extra blank line

sdmetrics/reports/single_table/diagnostic_report.py
```diff
@@ -1,75 +1,83 @@
...
+        Args:
+            property_name (str):
+                The name of the property.
+            metadata (dict):
+                The metadata of the table.
+        """
+        self._check_property_name(property_name)
+
+        num_columns = len(metadata['columns'])
+
```
pvk-developer, line 67: can remove this blank lines

sdmetrics/reports/single_table/diagnostic_report.py
```diff
@@ -1,75 +1,83 @@
...
+        if property_name in ['Coverage', 'Boundary']:
+            return num_columns
+        elif property_name == 'Synthesis':
```
pvk-developer, line 70: would use a blank line before this `elif` to see better the return

pvk-developer (changes requested):
Overall looks good! I just have some minor nits after that I think it should be good to merge 👍🏻

sdmetrics/reports/single_table/base_report.py
R-Palazzo: Yes that's a good point. @amontanez24, what would be the best here, making the method public, moving it directly to the `base_report.py` file or keeping it like this?

amontanez24: I think moving it to the BaseReport

sdmetrics/reports/single_table/diagnostic_report.py
amontanez24: If that's the case then we only need to run line 96 when verbose is False right?

sdmetrics/reports/single_table/diagnostic_report.py
```diff
@@ -49,31 +72,29 @@ def _get_num_iterations(self, property_name, metadata):
 
     def _print_results(self, out=sys.stdout):
```
amontanez24, line 73: unrelated but do we need the `out` parameter? Is it ever set to anything besides sys.stdout?

R-Palazzo: No that's a good point. Should I remove it?
Or modify the code to make the user able to use it (if they want to write the report output in a file for instance)

amontanez24: Yeah I think you can just remove it

sdmetrics/reports/single_table/base_report.py
R-Palazzo: Yes, done in [fcae5dc](https://github.com/sdv-dev/SDMetrics/pull/411/commits/fcae5dc0a63db277f9b06746c587f2fdd0fb2cd0).

sdmetrics/reports/single_table/base_report.py
R-Palazzo: Yes, done in [fcae5dc](https://github.com/sdv-dev/SDMetrics/pull/411/commits/fcae5dc0a63db277f9b06746c587f2fdd0fb2cd0).

sdmetrics/reports/single_table/diagnostic_report.py
R-Palazzo: Yes, done in [fcae5dc](https://github.com/sdv-dev/SDMetrics/pull/411/commits/fcae5dc0a63db277f9b06746c587f2fdd0fb2cd0).

sdmetrics/reports/single_table/diagnostic_report.py
R-Palazzo: Yes, done in [c277e80](https://github.com/sdv-dev/SDMetrics/pull/411/commits/c277e80072ca76014389bc3b1b12e6701d64747e).

pvk-developer (approved):
LGTM! :shipit:

sdmetrics/reports/single_table/diagnostic_report.py
R-Palazzo: Updated in [b12dcce](https://github.com/sdv-dev/SDMetrics/pull/411/commits/b12dccef1aa1a2ef01246e99ad8d4d804a3fc03c).

amontanez24 (approved):
Thanks for addressing the comments! :shipit:

PR #413, opened 2023-08-02
Update QualityReport properties return value

### LINKED ISSUE #405: The `Cardinality` property is not returning a DataFrame
### Environment Details
* SDMetrics version: Dev release in master branch (for upcoming SDMetrics 0.11.0)

### Error Description
In the new version of SDmetrics, we are renaming the `'Table Relationships'` property to `'Cardinality'` but it should otherwise behave the same. However, I am seeing that `get_details` for this property is returning a dictionary instead of a DataFrame, as expected.

### Steps to reproduce
Reproduce using the demo data.

```python
from sdmetrics.reports.multi_table import QualityReport

report = QualityReport()
report.generate(real_data, synthetic_data, metadata)
report.get_details(
    property_name='Cardinality',
    table_name='sessions'
)
```
Output:
```
{('sessions', 'transactions'): {'score': 0.7}}
```

(Note that one of the relationships is also missing from this output. But this is being covered in #404)

### Expected Behavior
Nothing should change from the older version except for the property name (`'Table Relationships'` --> `'Cardinality'`).  In the older version, the same method call returns this:

image

fealho:
Resolve #405.

### DIFF
--- a/sdmetrics/reports/multi_table/quality_report.py
+++ b/sdmetrics/reports/multi_table/quality_report.py
@@ -209,7 +209,7 @@ class QualityReport():
                 'Child Table': [tables[0]],
                 'Parent Table': [tables[1]],
                 'Metric': ['CardinalityShapeSimilariy'],
-                'Quality Score': [scores['score']]
+                'Score': [scores['score']]
             })
             details_dataframe = pd.concat([details_dataframe, new_row], ignore_index=True)
 
--- a/sdmetrics/reports/multi_table/quality_report.py
+++ b/sdmetrics/reports/multi_table/quality_report.py
@@ -176,6 +176,45 @@ class QualityReport():
 
         return self._properties_instances[property_name].get_visualization(table_name)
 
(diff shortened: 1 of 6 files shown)


sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -176,6 +176,45 @@ def get_visualization(self, property_name, table_name=None):
...
+                if table_name in table_names
+            }
+
+        details_dataframe = pd.DataFrame()
+        for tables, scores in details.items():
+            new_row = pd.DataFrame({
+                'Child Table': [tables[0]],
+                'Parent Table': [tables[1]],
+                'Metric': ['CardinalityShapeSimilariy'],
+                'Quality Score': [scores['score']]
```
amontanez24, line 212: I think we decided to just call it `Score`

amontanez24 (approved):
LGTM!

R-Palazzo (approved):
LGTM!

PR #416, opened 2023-08-09
Overall property score should be the average across all breakdowns

### LINKED ISSUE #415: Overall property score should be the average across all breakdowns
### Environment Details
* SDMetrics version: Dev version (master branch)

### Error Description
In the current development version, the overall property score is not being computed properly

- **Expected**: We expect this score to be the average all the individual breakdowns. Eg. the Column Shapes should be the average of all individual columns' scores
- **Actual**: We seem to be averaging across tables first, and then averaging the averages across tables. This biases the metric, because a table with fewer columns is being counted equally compared to a table with more columns.

### Steps to reproduce

```python
from sdmetrics.reports.multi_table import QualityReport

report = QualityReport()
```

Observe that the **Column Shapes** property score is reported as 79.68%. However, the average of Column Shapes is actually 79.23% -- we expect this number.

```python
all_shapes = report.get_details('Column Shapes')
print(all_shapes['Score'].agg('mean'))
```

amontanez24:
resolves #415

### DIFF
--- a/sdmetrics/reports/multi_table/_properties/base.py
+++ b/sdmetrics/reports/multi_table/_properties/base.py
@@ -1,5 +1,4 @@
 """Multi table base property class."""
-import numpy as np
 import pandas as pd
 
 
@@ -54,7 +53,7 @@ class BaseMultiTableProperty():
 
         self.is_computed = True
         all_details = pd.concat(all_details)
-        return np.nanmean(all_details['Score'])
+        return all_details['Score'].mean()
 
     def get_visualization(self, table_name):
         """Return a visualization for each score in the property.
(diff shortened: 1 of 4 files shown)


sdmetrics/reports/multi_table/_properties/base.py
```diff
@@ -39,17 +40,21 @@ def get_score(self, real_data, synthetic_data, metadata, progress_bar=None):
...
+                metadata['tables'][table_name],
                 progress_bar
             )
+            all_details.append(property_instance._details)
 
         self.is_computed = True
-
-        return np.nanmean(average_score)
+        all_details = pd.concat(all_details)
+        return np.nanmean(all_details['Score'])
```
R-Palazzo, line 57: We could use `all_details['Score'].mean()`

R-Palazzo (approved):
LGTM!

fealho (approved):

PR #421, opened 2023-08-18
Fix Single Table Quality Report rounding

### LINKED ISSUE #419: Single Table Quality Report is not rounding correctly
### Environment Details

* SDMetrics version: 0.11.0 (latest)

### Error Description
When generating the Quality Report with `verbose=True`, the report should display the final score and property scores rounded to the nearest 2 decimal digits. We did this to make it more readable to the end user.

In practice, there are some edge cases where it's unable to round to the nearest 2 decimal digits.

```
Generating report ...
(1/2) Evaluating Column Shapes: : 100%|██████████| 9/9 [00:00<00:00, 462.83it/s]
(2/2) Evaluating Column Pair Trends: : 100%|██████████| 36/36 [00:00<00:00, 81.39it/s]

Overall Quality Score: 87.18%

Properties:
- Column Shapes: 92.17%
- Column Pair Trends: 82.19999999999999%
```

### Steps to reproduce
I'm not sure if this will always trigger the case, but I first noticed it when running the SDV's CTGANSynthesizer demo.

```python
from sdv.single_table import CTGANSynthesizer
from sdv.evaluation.single_table import evaluate_quality

real_data, metadata = download_demo(
    modality='single_table',
    dataset_name='fake_hotel_guests'
)

synthesizer = CTGANSynthesizer(metadata)
synthesizer.fit(real_data)
synthetic_data = synthesizer.sample(num_rows=500)

quality_report = evaluate_quality(
    real_data,
    synthetic_data,
    metadata
)
```

### Additional Context
We are meant to be rounding the final score and the property scores similarly to the same # of digits. I think the multiplication by 100 is messing up the rounding. Best to do the rounding directly within the print statement.

https://github.com/sdv-dev/SDMetrics/blob/d34cae04e73bbeb4542454339fa9227bdde6fe09/sdmetrics/reports/single_table/quality_report.py#L76-L87

fealho:
Resolve #419.

### DIFF
--- a/sdmetrics/reports/single_table/quality_report.py
+++ b/sdmetrics/reports/single_table/quality_report.py
@@ -81,9 +81,9 @@ class QualityReport():
         out.write('Properties:\n')
 
         for property_name in self._properties:
-            property_score = round(self._properties[property_name]._compute_average(), 4)
+            property_score = round(self._properties[property_name]._compute_average() * 100, 2)
             out.write(
-                f'- {property_name}: {property_score * 100}%\n'
+                f'- {property_name}: {property_score}%\n'
             )
 
     def generate(self, real_data, synthetic_data, metadata, verbose=True):


R-Palazzo (approved):
LGTM! Have you tried with the example in the issue?

fealho:
@R-Palazzo Yep. Only thing that happens now is it returns a single digit sometimes (e.g. 10.8% instead of 10.80%), which I think is fine.

frances-h (approved):
👍

PR #436, opened 2023-08-29
RangeCoverage errors on datetime columns in DiagnosticReport

### LINKED ISSUE #431: RangeCoverage errors on datetime columns in DiagnosticReport
### Environment Details

* SDMetrics version: diagnostic-report-properties branch
* Python version: Any 
* Operating System: Any

### Error Description

When running the `DiagnosticReport`, datetime columns return an error for the `RangeCoverage` metric.

<div style="color: rgb(33, 33, 33); font-family: Roboto, Noto, sans-serif; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">

Column | Metric | Score | Error
-- | -- | -- | --
has_rewards | CategoryCoverage | 1.000000 | None
room_type | CategoryCoverage | 1.000000 | None
amenities_fee | RangeCoverage | 1.000000 | None
checkin_date | RangeCoverage | NaN | TypeError: unsupported operand type(s) for -: ...
checkout_date | RangeCoverage | NaN | TypeError: '<=' not supported between instance...
room_rate | RangeCoverage | 0.466742 | None

</div><br class="Apple-interchange-newline">

### Steps to reproduce

```python
from sdv.multi_table import HMASynthesizer

data, metadata = download_demo('multi_table', 'fake_hotels')
synth = HMASynthesizer(metadata)
synth.fit(data)
synthetic_data = synth.sample(len(data))

report = DiagnosticReport()
report.generate(data, synthetic_data, metadata.to_dict(), verbose=True)

report.get_details('Coverage', 'guests')
```

### Additional Context

- If the metadata says a column is datetime, but it is represented as a string, we should try to convert it to pd.datetime using the provided format in the report, and pass the converted column to all the metrics. If it can't be converted to a datetime object either because the format isn't provided or it just won't work, then we can just pass the column as is.

frances-h:
Resolve #431


tests/unit/reports/multi_table/test_multi_table_diagnostic_report.py
```diff
@@ -23,6 +23,61 @@ def test___init__(self):
         assert report._metric_averages == {}
         assert report._results == {}
 
+    @patch('sdmetrics.reports.multi_table.diagnostic_report.convert_datetime_columns')
+    def test__convert_datetimes(self, convert_datetime_mock):
+        """Test that ``_convert_datetimes`` tries to convert datetime columns."""
+        real_datetimes = pd.to_datetime(['2000-01-01', '2023-01-01', '2010-05-06'])
```
fealho, line 29: Missing # Setup

sdmetrics/reports/multi_table/diagnostic_report.py
```diff
@@ -73,6 +73,28 @@ def _print_results(self, out=sys.stdout):
         print_results_for_level(out, self._results, 'WARNING')
         print_results_for_level(out, self._results, 'DANGER')
 
+    def _convert_datetimes(self, real_data, synthetic_data, metadata):
```
fealho, line 76: I'd move this one to utils as well, and just call the single table `_convert_datetimes` for each table, so we avoid duplicated code.

fealho (commented):
Pretty good, just 1 small suggestion.

fealho (approved):
Thanks for addressing 👍

frances-h:
@fealho I realized that we probably want to convert datetimes across all reports, not just the `DiagnosticReport` so I moved it back into the `BaseReport`.

sdmetrics/reports/base_report.py
```diff
@@ -57,6 +59,28 @@ def validate(self, real_data, synthetic_data, metadata):
...
+        """
+        for column, col_meta in metadata['columns'].items():
+            if col_meta['sdtype'] == 'datetime':
+                real_col = real_data[column]
+                synth_col = synthetic_data[column]
+                try:
+                    converted_cols = convert_datetime_columns(real_col, synth_col, col_meta)
+                    real_data[column], synthetic_data[column] = converted_cols
+                except Exception:
+                    continue
```
pvk-developer, line 82: I believe that recently we decided to stop using `staticmethods` on classes, I would encourage to move this to a utility function and call it when needed.

sdmetrics/reports/multi_table/base_multi_table_report.py
```diff
@@ -47,6 +47,21 @@ def _check_table_names(self, table_name):
...
+        Args:
+            real_data (pandas.DataFrame):
+                The real data.
+            synthetic_data (pandas.DataFrame):
+                The synthetic data.
+            metadata (dict):
+                The metadata, which contains each column's data type as well as relationships.
+        """
+        for table, table_metadata in metadata['tables'].items():
+            BaseReport.convert_datetimes(real_data[table], synthetic_data[table], table_metadata)
```
pvk-developer, line 63: Instead of calling the base class static method use the utility function

pvk-developer (changes requested):

pvk-developer (approved):

sdmetrics/reports/base_report.py
pvk-developer: Oh I see now, feel free to merge 👍🏻

PR #437, opened 2023-08-30
Always show `Table` column in `get_details`

### LINKED ISSUE #429: Always show the `Table` column in `get_details`
### Environment Details
* SDMetrics version: In dev (`diagnostic-report-properties` branch)
* Python version: 3.10
* Operating System: Linux (Colab)

### Description
In a multi-table diagnostic report, the `table_name` parameter is optional when getting details. In the case where I don't provide a table name, the returned column (`Table`) is helpful in showing me which table the results are for.

In the case where I _do_ provide a table name, then we can keep the `Table` column for consistency's sake. (Even if it's always constant.) 

### Steps to reproduce

```python

report = DiagnosticReport()
report.get_details(property_name='Boundary', table_name='users')
```

Output: There is no column labeled `Table`
image

frances-h:
Resolve #429


amontanez24 (approved):
:shipit:

fealho (approved):

PR #438, opened 2023-08-31
Diagnostic explanations should not repeat if I generate multiple times

### LINKED ISSUE #430: Diagnostic explanations should not repeat if I generate multiple times
### Environment Details
* SDMetrics version: In dev (`diagnostic-report-properties` branch)
* Python version: 3.10
* Operating System: Linux (Colab)

### Error Description
If I generate the same DiagnosticReport object multiple times, the printed out results repeat and accumulate during every run. We should make sure they are cleared every time we start to generate.

### Steps to reproduce

This is just a toy example I was working with.

```python
import pandas as pd
from sdmetrics.reports.single_table import DiagnosticReport

my_data = pd.DataFrame(data={
    'A': ['a', 'b', 'c'],
    'B': ['e', 'f', 'f']
})

my_data2 = pd.DataFrame(data={
    'A': ['b', 'c', 'a'],
    'B': ['f', 'f', 'e']
})

metadata = {
    'columns': {
      'A': { 'sdtype':  'address' },
      'B': { 'sdtype': 'phone_number'}
    }
}

report = DiagnosticReport()
report.generate(my_data, my_data2, metadata, verbose=False)
report.generate(my_data, my_data2, metadata)
```

Output (See how the DANGER message is repeating)
```
Generating report ...
(1/3) Evaluating Coverage: : 100%|██████████| 2/2 [00:00<00:00, 217.24it/s]
(2/3) Evaluating Boundary: : 100%|██████████| 2/2 [00:00<00:00, 199.55it/s]
(3/3) Evaluating Synthesis: : 100%|██████████| 1/1 [00:00<00:00,  7.56it/s]

Diagnostic Results:

DANGER:
x More than 50% of the synthetic rows are copies of the real data
x More than 50% of the synthetic rows are copies of the real data
```

amontanez24:
resolves #430 

This PR makes a few structural changes. The problem is that there needed to be a `MultiTableBaseReport` because the method signatures are different for a lot of the methods. So inheritance based on the number of tables made sense. However, the displaying/printing of results was the same for report types. In other words, both `multi_table.DiagnosticReport` and `single_table.DiagnosticReport` print the exact same way. Same thing for the quality reports. To handle this, a lot of the duplicate code was moved into the [utils.py](https://github.com/sdv-dev/SDMetrics/blob/577aab63f3a029ae387ac3c00a878a66cf819006/sdmetrics/reports/utils.py#L829). This was awkward because it had utility functions actually changing the objects themselves.

The root cause of the problem in this issue is that we wanted to generate the details in a diagnostic report even if `verbose` was False, but we only wanted to print it if it was True. Since the code for generating the details was put in the printing function for reusability, it created some strange dependencies. The code was calling the [_generate_details ](https://github.com/sdv-dev/SDMetrics/blob/577aab63f3a029ae387ac3c00a878a66cf819006/sdmetrics/reports/utils.py#L829) method twice in the case where `verbose` was True. To not rerun code, the following lines were added
https://github.com/sdv-dev/SDMetrics/blob/577aab63f3a029ae387ac3c00a878a66cf819006/sdmetrics/reports/utils.py#L831-L834
The problem is that they prevented the details from clearing whenever the method was called more than once consecutively.

To resolve this issue and the awkward dependencies I proposed moving all of the printing functionality to a class called a `ResultsHandler`. Each report points to the type of `ResultsHandler` it needs and that class handles all the printing. It allows for the code to be reused without having external files modify attributes of other class objects. This could also have been achieved using a mixin but I thought this was simpler.

Additionally, this PR 
- Renames some test files to match the actual file names
- Makes the `_details` attribute of the single table property's public to match the multi table properties.


frances-h (approved):
Looks good!

PR #446, opened 2023-09-13
Fix convert datetimes in BaseReport

### LINKED ISSUE #431: RangeCoverage errors on datetime columns in DiagnosticReport

frances-h:
Resolve #431 

- Add missing line that was deleted from #436 
- Add tests to test object datetimes in reports

### DIFF
--- a/tests/unit/reports/test_base_report.py
+++ b/tests/unit/reports/test_base_report.py
@@ -179,8 +179,8 @@ class TestBaseReport:
         })
         metadata = {
             'columns': {
-                'column1': {'sdtypes': 'numerical'},
-                'column2': {'sdtypes': 'categorical'}
+                'column1': {'sdtype': 'numerical'},
+                'column2': {'sdtype': 'categorical'}
             }
         }
 
@@ -238,10 +238,10 @@ class TestBaseReport:
         })
         metadata = {
             'columns': {
-                'column1': {'sdtypes': 'numerical'},
-                'column2': {'sdtypes': 'categorical'},
-                'column3': {'sdtypes': 'numerical'},
-                'column4': {'sdtypes': 'numerical'},
+                'column1': {'sdtype': 'numerical'},
+                'column2': {'sdtype': 'categorical'},
+                'column3': {'sdtype': 'numerical'},
+                'column4': {'sdtype': 'numerical'},
             }
         }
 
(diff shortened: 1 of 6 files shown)


R-Palazzo (approved):
LGTM!

tests/unit/reports/test_base_report.py
```diff
@@ -179,8 +179,8 @@ def test_generate(self):
         })
         metadata = {
             'columns': {
-                'column1': {'sdtypes': 'numerical'},
-                'column2': {'sdtypes': 'categorical'}
+                'column1': {'sdtype': 'numerical'},
```
amontanez24, line 182: good catch!

amontanez24 (approved):
:shipit:

PR #455, opened 2023-09-25
Add `get_column_plot`

### LINKED ISSUE #443: Add new get_column_plot
### Problem Description

_As a user, sometimes the plot type selected by `get_column_plot` is not the best. I'd like to be able to select the plot type to better visualize my data._

### Expected behavior

- In the `sdmetrics.visualization` module, add a method called `get_column_plot`.
    - If the module/file doesn't exist, create it.
- Use the current [get_column_plot ](https://github.com/sdv-dev/SDMetrics/blob/e0a23bfd99d839f520055d9b74498c9d75644b84/sdmetrics/reports/utils.py#L271C13-L271C13)function as a starting point.
    - Remove the metadata parameter
    - Add a new parameter called `plot_type`
        - `plot_type` can be 'distplot', 'bar' or None
        - By default, use the dtype of the column to pick. If the dtype is datetime, int or floats, use 'displot'. Otherwise use 'bar'.
- Delete the old `get_column_plot` code.

### Additional context
- If provided, verify that a valid plot_type is given.
- We no longer need any of the metadata validation

pvk-developer:
Resolves #443 

image
image
image
image


frances-h (approved):
Looks good! Thanks for catching that `nan` bug.

sdmetrics/visualization.py
```diff
@@ -259,3 +263,49 @@ def get_cardinality_plot(real_data, synthetic_data, child_table_name, parent_tab
...
+    Args:
+        real_data (pandas.DataFrame):
+            The real table data.
+        synthetic_data (pandas.DataFrame):
+            The synthetic table data.
+        column_name (str):
+            The name of the column.
+        plot_type (str or None):
+            The plot to be used. Can choose between ``distplot``, ``bar`` or ``None``. If ``None`
+            select between ``distplot`` or ``bar`` depending on the data that the column contains,
```
amontanez24, line 280: minor: can we put a `:` after contains instead of a comma

sdmetrics/visualization.py
```diff
@@ -259,3 +263,49 @@ def get_cardinality_plot(real_data, synthetic_data, child_table_name, parent_tab
...
+        )
+
+    if column_name not in real_data.columns:
+        raise ValueError(f"Column '{column_name}' not found in real table data.")
+    if column_name not in synthetic_data.columns:
+        raise ValueError(f"Column '{column_name}' not found in synthetic table data.")
+
+    column_is_datetime = is_datetime(real_data[column_name])
+    real_column = real_data[column_name]
+    dtype = real_column.dropna().infer_objects().dtype.kind
```
amontanez24, line 299: minor: I think `column_is_datetime` and `dtype` can be moved into the `if plot_type is None` block since they're only used in that case.

amontanez24 (approved):
A couple minor comments but after that 🚢

amontanez24:
also can we rename the PR to `get` instead of `generate`

PR #456, opened 2023-09-25
Add `InterTableTrends` property

### LINKED ISSUE #451: Add InterTable Trends property
### Problem Description

_As a user, I'd like to have information on how well my synthesized data maintained relationships between different tables._

### Expected behavior

- Add a new multi table property class called `InterTableTrends`.
- Add this property to the multi table quality report

#### Score calculation

This property is similar in principle to the `ColumnPairTrends` property, except it compares columns in different tables. In order to do this, we propose the following procedure:
1. For every relationship listed in the metadata, the property should merge on the primary/foreign keys in order to create a denormalized table. We do not need to merge any non-statistical columns (PII).
    1. _Note that in the denormalized table, the parent table's values will repeat. For example since one user has many transactions, the user's info will be repeated for each transaction._

user.id | user.age | user.gender | transaction.amt | transaction.date | ...
-- | -- | -- | -- | -- | --
0 | 22 | F | 23.50 | 2/12/23 | …
0 | 22 | F | 12.50 | 4/29/22 | …
0 | 22 | F | 100.00 | 8/31/23 | …
1 | 45 | M | 50.50 | 3/21/22 | …
1 | 45 | M | 500.00 | 8/23/22 | …

2. On the denormalized dataset, we can then compute the Column Pair Trends as we would on a single table.
    1. Notable exception: Do NOT compute Column Pair Trends for 2 columns that come from the same table (eg. user.age and user.gender). Because we will already have results for this!
3. The property score will be the average of all these scores.

**Note:** Performance is important! Ideally, the addition of this property should not add significant time to the computation. If it does, then we may want to consider subsetting the data that we use for the computation.

#### Methods

- `get_score(self, real_data, synthetic_data, metadata, progress_bar=None)`: Return the overall score.
    - This method is probably where the logic above will go.
- `get_details(table_name=None)`: This will return a dataframe with all the details. It should have the following columns:
    - Parent Table - Name of the parent table.
    - Child Table - Name of the child table.
    - Foreign Key - The name of the foreign key column.
    - Column 1 - Name of column in the parent table being compared.
    - Column 2 - Name of column in the child table being compared.
    - Metric - The metric to compare the columns.
    - Score - The score.
    - Real Correlation - The correlation in the real data.
    - Synthetic Correlation - The correlation in the synthetic data.
- `get_visualization(table_name=None)`: Create a simple bar graph that shows the final score between every pair of columns (similar to the bar graph for Column Shapes).
<html>
<body>

</body>
</html>

- Label the bars with the names of the table(s) as well as the names of the columns
Eg. "user.age, transaction.amt"
- The color of the bars can signify which metric we used, similar to Column Shapes
- On hover, there should be more details about the correlations. It should be like a tooltip with the following information.
```
user.age, transaction.amt
Foreign Key: user_id


Metric=CorrelationSimilarity
Score=0.28123123
Real Correlation=0.5845345
Synthetic Correlation=0.345345123
```
- Note that `table_name` is required, but we need to consider all relationships that table is involved with (either a parent or as a child).

### Additional context

- The new property should also have a progress bar that increments by 1 every time we compute a pairwise trend.
    - Either the base `_get_num_iterations` will have to change or it will need to be overridden in this class because none of the current options match this scenario.
https://github.com/sdv-dev/SDMetrics/blob/3667b377e4fb38c745895d7f35b3a3a41a7cf902/sdmetrics/reports/multi_table/_properties/base.py#L29-L39

frances-h:
Resolve #451

### DIFF
--- a/sdmetrics/reports/multi_table/_properties/inter_table_trends.py
+++ b/sdmetrics/reports/multi_table/_properties/inter_table_trends.py
@@ -1,0 +1,172 @@
+"""Column pair trends property for multi-table."""
+import itertools
+
+import numpy as np
+import pandas as pd
+import plotly.express as px
+
+from sdmetrics.reports.multi_table._properties import BaseMultiTableProperty
+from sdmetrics.reports.single_table._properties import (
+    ColumnPairTrends as SingleTableColumnPairTrends)
+from sdmetrics.reports.utils import PlotConfig
+
+
+class InterTableTrends(BaseMultiTableProperty):
+    """Column pair trends property for multi-table.
+
+    This property evaluates the matching in trends between pairs of real
+    and synthetic data columns across related tables. Each pair's correlation is
+    calculated and the final score represents the average of these measures across
+    all column pairs
+    """
+    _num_iteration_case = 'inter_table_columns'
+
+    def get_score(self, real_data, synthetic_data, metadata, progress_bar=None):
+        """Get the average score of all the individual metric scores computed.
+
+        Args:
+            real_data (pandas.DataFrame):
+                The real data.
+            synthetic_data (pandas.DataFrame):
+                The synthetic data.
+            metadata (dict):
+                The metadata, which contains each column's data type as well as relationships.
+            progress_bar (tqdm.tqdm or None):
+                The progress bar object. Defaults to None.
+
+        Returns:
+            float:
+                The average score for the property for all the individual metric scores computed.
+        """
+        all_details = pd.DataFrame(columns=[
+            'Parent Table',
+            'Child Table',
+            'Foreign Key',
+            'Column 1',
+            'Column 2',
+            'Metric',
+            'Score',
+            'Real Correlation',
+            'Synthetic Correlation',
+            'Error'
+        ])
+        for relationship in metadata['relationships']:
+            parent = relationship['parent_table_name']
+            child = relationship['child_table_name']
+            foreign_key = relationship['child_foreign_key']
+            primary_key = relationship['parent_primary_key']
+
+            real_parent = real_data[parent].add_prefix(f'{parent}.')
+            real_child = real_data[child].add_prefix(f'{child}.')
+            synthetic_parent = synthetic_data[parent].add_prefix(f'{parent}.')
+            synthetic_child = synthetic_data[child].add_prefix(f'{child}.')
+
+            child_index = f'{child}.{foreign_key}'
+            parent_index = f'{parent}.{primary_key}'
+
+            denormalized_real = real_child.merge(
+                real_parent, left_on=child_index, right_on=parent_index
+            )
+            denormalized_synthetic = synthetic_child.merge(
+                synthetic_parent, left_on=child_index, right_on=parent_index
+            )
+
+            child_meta = metadata['tables'][child]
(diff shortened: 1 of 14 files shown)


tests/integration/reports/multi_table/_properties/test_inter_table_trends.py
```diff
@@ -0,0 +1,43 @@
...
+        assert result == 0.48240740740740734
+
+    def test_with_progress_bar(self):
+        """Test that the progress bar is correctly updated."""
+        # Setup
+        real_data, synthetic_data, metadata = load_demo(modality='multi_table')
+        column_pair_trends = InterTableTrends()
+        num_iter = sum(
+            len(metadata['tables'][relationship['parent_table_name']]['columns'])
+            * len(metadata['tables'][relationship['child_table_name']]['columns'])  # noqa: W503
```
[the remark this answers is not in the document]
frances-h: Moved ignoring the rule to setup.cfg

sdmetrics/reports/multi_table/_properties/inter_table_trends.py
```diff
@@ -0,0 +1,197 @@
...
+            merged_metadata = metadata['tables'][child].copy()
+            child_cols = {
+                f'{child}.{col}': col_meta for col, col_meta in child_meta['columns'].items()
+            }
+            parent_cols = {
+                f'{parent}.{col}': col_meta for col, col_meta in parent_meta['columns'].items()
+            }
+            merged_metadata['columns'] = {**child_cols, **parent_cols}
+            if 'primary_key' in merged_metadata:
+                merged_metadata['primary_key'] = f'{child}.{merged_metadata["primary_key"]}'
```
amontanez24, line 86: might be helpful to pull into helper method

sdmetrics/reports/multi_table/_properties/inter_table_trends.py
```diff
@@ -0,0 +1,197 @@
...
+            details['Foreign Key'] = foreign_key
+            details['Column 1'] = details['Column 1'].str.replace(f'{parent}.', '', n=1)
+            details['Column 2'] = details['Column 2'].str.replace(f'{child}.', '', n=1)
+
+            all_details = pd.concat([all_details, details]).reset_index(drop=True)
+
+        if all_details['Error'].isna().all():
+            all_details = all_details.drop('Error', axis=1)
+        else:
+            all_details['Error'] = all_details['Error'].replace({np.nan: None})
```
amontanez24, line 106: wondering if we can pull this out somehow since it seems like we do it in every property

amontanez24 (commented):
I think the approach seems reasonable!

R-Palazzo (approved):
LGTM! Thanks for the cleaning ;)

sdmetrics/reports/multi_table/_properties/inter_table_trends.py
frances-h: I ended up refactoring the base `.get_score` method to handle this. Now it will call `_generate_details` to create the details dataframe, then clean it up and return the average score.

amontanez24 (approved):
LGTM!

R-Palazzo (approved):

PR #457, opened 2023-09-25
Add `get_column_pair_plot`

### LINKED ISSUE #444: Add new get_column_pair_plot
### Problem Description

_As a user, sometimes the plot type selected by get_column_pair_plot is not the best. I'd like to be able to select the plot type to better visualize my data._

### Expected behavior

- In the `sdmetrics.visualization` module, add a method called `get_column_pair_plot`.
    - If the module/file doesn't exist, create it.
- Use the current [get_column_pair_plot ](https://github.com/sdv-dev/SDMetrics/blob/e0a23bfd99d839f520055d9b74498c9d75644b84/sdmetrics/reports/utils.py#L435) function as a starting point.
    - Remove the metadata parameter
    - Add a new parameter called `plot_type`
        - `plot_type` can be 'scatter', 'box', 'heatmap' or None
        - By default, use the dtype of the columns to pick.
            - If both are float/int/datetime, then use 'scatter'
            - If one is float/int/datetime and the other isn't use 'box'
            - If neither are float/int/datetime, use 'heatmap'
- Delete the old `get_column_pair_plot` code.

### Additional context

- If provided, verify that a valid plot_type is given.
- We no longer need any of the metadata validation

pvk-developer:
Resolves #444 

image
image
image
image


sdmetrics/visualization.py
```diff
@@ -309,3 +426,67 @@ def get_column_plot(real_data, synthetic_data, column_name, plot_type=None):
...
+        plot_type (str or None):
+            The plot to be used. Can choose between ``box``, ``heatmap``, ``scatter`` or ``None``.
+            If ``None` select between ``box``, ``heatmap`` or ``scatter`` depending on the data
+            that the column contains, ``scatter`` used for datetime and numerical values,
+            ``heatmap`` for categorical and ``box`` for a mix of both. Defaults to ``None``.
+
+    Returns:
+        plotly.graph_objects._figure.Figure
+    """
+    if len(column_names) != 2:
```
R-Palazzo, line 450: Maybe we can put all the validation in a separate private method, like `_validate_input_plot()` or something else

tests/unit/test_visualization.py
```diff
@@ -363,3 +364,288 @@ def test_get_column_plot_plot_type_distplot(mock__generate_column_plot):
...
+    with pytest.raises(ValueError, match=match):
+        get_column_pair_plot(real_data, synthetic_data, columns)
+
+
+def test_get_column_pair_plot_invalid_plot_type():
+    """Test when invalid plot type is passed as argument to ``get_column_pair_plot``."""
+    # Setup
+    columns = ['start_date', 'end_date']
+    real_data = synthetic_data = pd.DataFrame({'start_date': [], 'end_date': []})
+    # Run and Assert
```
R-Palazzo, line 527: add blank line

R-Palazzo (commented):
Looks good! Thanks for the cleaning and the visualizations.
I realized that when you switch the column order, it changes the boxplot orientation.
Maybe that's something we can change here, always to have the categorical column on the` x-axis` and the continuous column on the` y-axis` (or the reverse). @npatki does it make sense?

sdmetrics/visualization.py
```diff
@@ -34,9 +34,127 @@ def _generate_column_bar_plot(real_data, synthetic_data, plot_kwargs={}):
...
+
+    Returns:
+        plotly.graph_objects._figure.Figure
+    """
+    columns = real_data.columns
+    real_data = real_data.copy()
+    real_data['Data'] = 'Real'
+    synthetic_data = synthetic_data.copy()
+    synthetic_data['Data'] = 'Synthetic'
+    all_data = pd.concat([real_data, synthetic_data], axis=0, ignore_index=True)
```
amontanez24, line 63: This part seems to be the same for all the plots. Maybe it should be done in `get_column_pair_plot` instead?

sdmetrics/visualization.py
```diff
@@ -105,25 +223,28 @@ def _generate_column_plot(real_column,
...
+        real_data['values'] = real_data['values'].astype('int64')
+        synthetic_data['values'] = synthetic_data['values'].astype('int64')
 
     trace_args = {}
 
     if plot_type == 'bar':
         fig = _generate_column_bar_plot(real_data, synthetic_data, plot_kwargs)
     elif plot_type == 'distplot':
+        if x_label is None:
+            x_label = 'Value'
```
amontanez24, line 246: I think you can do this on one line

amontanez24 (approved):
Looks good! Just a couple minor comments

npatki:
> I realized that when you switch the column order, it changes the boxplot orientation. Maybe that's something we can change here, always to have the categorical column on the` x-axis` and the continuous column on the` y-axis` (or the reverse). @npatki does it make sense?

For now, this is working as intended. We weren't sure which people would like more, and it's easy to switch the column order (in the input) to get the orientation you want.

R-Palazzo (approved):
LGTM!

frances-h (approved):

PR #459, opened 2023-10-02
Expose reports module in top-level init

frances-h:
In order for add-ons to work correctly with the reports, we need to expose them to the top-level __init__.


rwedge (approved):

pvk-developer (approved):
LGTM!

PR #466, opened 2023-10-12
Validate that the metadata is always a dict

### LINKED ISSUE #428: Validate that the metadata is always a dict
### Problem Description
The diagnostic and quality reports always expect that the metadata is a _dictionary_. This can be confusing because in other libraries (such as the SDV), the metadata is actually its own object.

### Expected behavior
If I don't pass in a dictionary into a report, provide a helpful error message.

```python
from sdmetrics.reports.single_table import DiagnosticReport

report = DiagnosticReport()
report.generate(real_data, synthetic_data, metadata_object)
```

```
InvalidInputError: The provided metadata is not a dictionary.
```

### Additional context
We cannot use metadata objects in SDMetrics because we are not using or importing the SDV.

R-Palazzo:
Resolve #428

### DIFF
--- a/sdmetrics/reports/base_report.py
+++ b/sdmetrics/reports/base_report.py
@@ -55,13 +55,24 @@ class BaseReport():
             metadata (dict):
                 The metadata of the table.
         """
-        if not isinstance(metadata, dict):
-            metadata = metadata.to_dict()
-
         self._validate_metadata_matches_data(real_data, synthetic_data, metadata)
 
     def _handle_results(self, verbose):
         raise NotImplementedError
+
+    @staticmethod
+    def _convert_metadata(metadata):
+        """If the metadta is not a dict, try to convert it."""
+        if not isinstance(metadata, dict):
+            try:
+                metadata = metadata.to_dict()
+            except Exception:
+                raise TypeError(
+                    'The provided metadata is not a dictionary and does not have a to_dict method.'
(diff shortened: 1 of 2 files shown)


npatki:
@R-Palazzo we should not be accepting the actual metadata object. We should only accept dictionaries.

I would avoid doing any conversions in SDMetrics and stick to the original issue.

sdmetrics/reports/base_report.py
```diff
@@ -55,14 +55,25 @@ def validate(self, real_data, synthetic_data, metadata):
...
 
     def _handle_results(self, verbose):
         raise NotImplementedError
 
+    @staticmethod
+    def _convert_metadata(metadata):
+        """If the metadta is not a dict, try to convert it."""
+        if not isinstance(metadata, dict):
+            try:
+                metadata = metadata.to_dict()
```
frances-h, line 68: Is this something we actually want to support? Other objects might implement a `to_dict` method which might lead to weird issues down the road (i.e. I'm pretty sure you can call `to_dict` on a pandas dataframe). Maybe @npatki has thoughts?

npatki: @frances-h agreed. I think we should stick to the issue's description, which states that we should only accept dictionaries. We do not intend to convert to a dictionary on the user's behalf.

amontanez24 (approved):
LGTM!

PR #474, opened 2023-10-23
Add `KeyUniqueness` metric

### LINKED ISSUE #460: Add `KeyUniqueness` metric
### Problem Description
_As a user, I would like a metric that gives me information about the uniqueness of my primary key and alternate key columns._

### Expected behavior
- Add a new `single_column` metric that calculates the percent of keys that are unique and not null.
- This metric takes in primary key or alternate key columns (either ID or PII sdtypes).

#### Attributes
- `name`: `'KeyUniqueness'`
- `goal`: `Goal.MAXIMIZE`
- `min_value`: 0.0
- `max_value`: 1.0

#### Methods
- `compute(real_data, synthetic_data)`: Compute the score for the metric. The returned score should be the percent of keys that are unique and not null (eg. a score of 0.6 means 60% of the keys are unique and 40% are duplicates).  
    - If the real data does not pass this test, then the metric should Error.

R-Palazzo:
Clickup task: CU-86ayfczkn
Resolve #460

### DIFF
--- a/sdmetrics/single_column/statistical/key_uniqueness.py
+++ b/sdmetrics/single_column/statistical/key_uniqueness.py
@@ -1,0 +1,80 @@
+"""Key Uniqueness Metric."""
+
+from sdmetrics.errors import InvalidDataError
+from sdmetrics.goal import Goal
+from sdmetrics.single_column.base import SingleColumnMetric
+
+
+class KeyUniqueness(SingleColumnMetric):
+    """Key uniqueness metric.
+
+    Compute the fraction of rows in the synthetic data that are unique.
+
+    Attributes:
+        name (str):
+            Name to use when reports about this metric are printed.
+        goal (sdmetrics.goal.Goal):
+            The goal of this metric.
+        min_value (Union[float, tuple[float]]):
+            Minimum value or values that this metric can take.
+        max_value (Union[float, tuple[float]]):
+            Maximum value or values that this metric can take.
+    """
+
+    name = 'KeyUniqueness'
+    goal = Goal.MAXIMIZE
+    min_value = 0.0
+    max_value = 1.0
+
+    @classmethod
+    def compute_breakdown(cls, real_data, synthetic_data):
+        """Compute the score breakdown of the key uniqueness metric.
+
+        Args:
+            real_data (pandas.Series):
+                The real data.
+            synthetic_data (pandas.Series):
+                The synthetic data.
+
+        Returns:
+            dict:
+                The score breakdown of the key uniqueness metric.
+        """
+        has_duplicates = real_data.duplicated().any()
+        if has_duplicates:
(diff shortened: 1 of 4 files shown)


sdmetrics/single_column/statistical/key_uniqueness.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Returns:
+            float:
+                The proportion of data points in the synthetic data that are unique.
+        """
+        return cls.compute_breakdown(real_data, synthetic_data)['score']
+
+    @classmethod
+    def normalize(cls, raw_score):
+        """Return the `raw_score` as is, since it is already normalized.
```
rwedge, line 73: Based on this description I would expect `return raw_score` to be the function logic

sdmetrics/single_column/statistical/key_uniqueness.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Args:
+            raw_score (float):
+                The value of the metric from `compute`.
+
+        Returns:
+            float:
+                The normalized value of the metric
+        """
+        return super().normalize(raw_score)
```
frances-h, line 83: Since we're just calling the parent class's `normalize`, we don't need to re-define it on the child class.

R-Palazzo: Yes good point, addressed in [5fdb9f5](https://github.com/sdv-dev/SDMetrics/pull/474/commits/5fdb9f5ae1b54d2c18fdb32b8d925664f653f294)

sdmetrics/single_column/statistical/key_uniqueness.py
R-Palazzo: Yes, changed in [5fdb9f5](https://github.com/sdv-dev/SDMetrics/pull/474/commits/5fdb9f5ae1b54d2c18fdb32b8d925664f653f294)

sdmetrics/single_column/statistical/key_uniqueness.py
```diff
@@ -0,0 +1,69 @@
...
+                The synthetic data.
+
+        Returns:
+            dict:
+                The score breakdown of the key uniqueness metric.
+        """
+        has_duplicates = real_data.duplicated().any()
+        has_nans = real_data.isna().any()
+        if has_duplicates or has_nans:
+            LOGGER.info('The real data contains NA or duplicate values.')
```
rwedge, line 48: should this be a `ValueError` or some other error type?

R-Palazzo: This is something we discussed yesterday at the Eng meeting. We decided to use a `logger.info()` to be consistent with a similar behavior in the `ReferentialIntegrity`.

rwedge (approved):

frances-h (approved):

PR #478, opened 2023-10-24
Fix `BoundaryAdherence` NaN handling

### LINKED ISSUE #470: Fix BoundaryAdherence NaN handling
### Environment Details

* SDMetrics version: 0.11.0 (latest)
* Python version: 3.10
* Operating System: MacOS

### Error Description
Currently, the `BoundaryAdherence` metric ignores NaN values when calculating the percentage of values that fall within the min/max range of the real values. NaN values in the synthetic data should only be ignored if the real data contains NaNs, otherwise they should be considered out of bounds.

### Proposed Fix
When computing, only drop NaN values from the synthetic data if there are NaN values present in the real data.

### Steps to reproduce
```python
import pandas as pd
import numpy as np

from sdmetrics.single_column import BoundaryAdherence

real_data = pd.Series([2.3, 4.2, 1.0, 10.0, 9.2])
synthetic_data = pd.Series([3.4, np.nan, 1.2, 4.9, 9.1])

BoundaryAdherence.compute(real_data, synthetic_data)
```

frances-h:
CU-86ayga3rk

Resolve #470

### DIFF
--- a/tests/unit/single_column/statistical/test_boundary_adherence.py
+++ b/tests/unit/single_column/statistical/test_boundary_adherence.py
@@ -1,5 +1,7 @@
+from datetime import datetime
 from unittest.mock import patch
 
+import numpy as np
 import pandas as pd
 
 from sdmetrics.single_column.statistical import BoundaryAdherence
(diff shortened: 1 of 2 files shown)


lajohn4747 (approved):
lgtm

tests/unit/single_column/statistical/test_boundary_adherence.py
```diff
@@ -31,6 +33,65 @@ def test_compute(self):
...
+        out of bounds if the real data does not also containt nan values.
+        """
+        # Setup
+        real_data = pd.Series([
+            datetime(2020, 10, 1),
+            datetime(2021, 1, 2),
+            datetime(2021, 9, 12),
+            datetime(2022, 10, 1),
+
+        ], dtype='datetime64[ns]')  # 0.8 -> 2.6
```
R-Palazzo, line 70: What does indicate # 0.8 -> 2.6 in the datetime case?

R-Palazzo (approved):
LGTM!

tests/unit/single_column/statistical/test_boundary_adherence.py
frances-h: Ah good catch, that was just to help me write the test 😅

PR #483, opened 2023-10-26
Add `Structure` property

### LINKED ISSUE #468: Add `Structure` property
### Problem Description
_As a user, I'd like a property that checks to see whether the overall structure of the synthetic data matches the real data_

Once we create the TableFormat metric (#463), we'd like to use it to calculate this new property.

### Expected behavior
- Add a new property `Structure` for single and multi table reports.
- The multi-table property can just call the single table property for each table.

#### Attributes
The property should have the following attributes:
- `_num_iteration_case = 'table'`

#### Score Calculation
This property should use the `TableFormat` metric to calculate the score for each table. The metric should ignore checking the dtype for ID and PII columns by passing them to the metric using the `ignore_dtype_columns` parameter. 

#### Methods
- `_generate_details(self, real_data, synthetic_data, metadata, progress_bar=None)`: This method will generate the `details` dataframe for the property. 
        - `Metric`: The metric that is applied to the given column
    - The progress bar should update for the table has been evaluated.
- `get_visualization(self)`: This method should return a bar graph of the scores for each column, similar to other properties.
    - Label the bars using the column name
    - On hover, there should be more details about the metric applied and the score

R-Palazzo:
CU-86ayg9qwy
Resolve #468

Let me know if this visualization works @frances-h, @npatki
Screenshot 2023-10-26 at 08 23 20

### DIFF
--- a/sdmetrics/reports/single_table/_properties/structure.py
+++ b/sdmetrics/reports/single_table/_properties/structure.py
@@ -79,7 +79,7 @@ class Structure(BaseSingleTableProperty):
         fig = px.bar(
             data_frame=self.details.dropna(subset=['Score']),
             y='Score',
-            title=f'Data Diagnostics: Structure (Average Score={round(average_score, 2)})',
+            title=f'Data Diagnostic: Structure (Average Score={round(average_score, 2)})',
             color='Metric',
             pattern_shape='Metric',
         )
--- a/sdmetrics/reports/single_table/_properties/structure.py
+++ b/sdmetrics/reports/single_table/_properties/structure.py
@@ -1,0 +1,104 @@
+import numpy as np
+import pandas as pd
+import plotly.express as px
+
+from sdmetrics.reports.single_table._properties import BaseSingleTableProperty
+from sdmetrics.single_table import TableFormat
+
+
+class Structure(BaseSingleTableProperty):
+    """Structure property class for single table.
+
+    This property checks to see whether the overall structure of the synthetic
+    data is the same as the real data.
+    """
+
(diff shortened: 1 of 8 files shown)


npatki:
@R-Palazzo seems OK for an MVP.

sdmetrics/reports/single_table/_properties/structure.py
```diff
@@ -0,0 +1,96 @@
...
+
+        Returns:
+            plotly.graph_objects._figure.Figure
+        """
+        average_score = self._compute_average()
+
+        fig = px.bar(
+            data_frame=self.details.dropna(subset=['Score']),
+            y='Score',
+            title=f'Data Diagnostics: Structure (Average Score={round(average_score, 2)})',
```
npatki, line 82: @R-Palazzo Title should be **Data Diagnostic** (singular, not plural)

We should make sure the title of these visualizations are consistent everywhere

npatki (changes requested):
@R-Palazzo one small title change. I also wonder if you can share a screenshot of the tooltip (on-hover) to verify that it makes sense for the visualization.

sdmetrics/reports/single_table/_properties/structure.py
R-Palazzo: Good catch, fixed in [620285f](https://github.com/sdv-dev/SDMetrics/pull/483/commits/620285f1f92ec9e57d6e5128d32d2cbabf16afbe)

frances-h (approved):

fealho (approved):
Looks pretty good 👍

PR #484, opened 2023-10-26
Add CardinalityBoundaryAdherence metric

### LINKED ISSUE #464: Add `CardinalityRuleMatch` metric
### Problem Description
_As a user, I would like a metric that checks the cardinality of my data falls within the expected range._

### Expected behavior
- Add a new `column_pair` metric that calculates the percent of synthetic parents whose cardinality falls within the minimum/maximum range of cardinality in the real data. 

#### Attributes
- `name`: `'CardinalityRuleMatch'`
- `goal`: `Goal.MAXIMIZE`
- `min_value`: 0.0
- `max_value`: 1.0

#### Methods
- `compute(real_data, synthetic_data)`: Compute the score for the metric. The returned score should be the percent of synthetic parents which have a number of children in the expected range. Eg. if the real data has parents all with at least 1 child and at most 4 children, the score would be the percent of synthetic parents with at least 1 child and no more than 4 children.

frances-h:
CU-86ayfd8dr

Resolve #464

### DIFF
--- a/sdmetrics/column_pairs/statistical/cardinality_boundary_adherence.py
+++ b/sdmetrics/column_pairs/statistical/cardinality_boundary_adherence.py
@@ -1,0 +1,80 @@
+"""ColumnPair metrics based on Kullback–Leibler Divergence."""
+
+import pandas as pd
+
+from sdmetrics.column_pairs.base import ColumnPairsMetric
+from sdmetrics.goal import Goal
+
+
+class CardinalityBoundaryAdherence(ColumnPairsMetric):
+    """Cardinality Boundary Adherence metric.
+
+    Computes the percentage of synthetic parents whose cardinality
+    falls within the min/max range of cardinality in the real data.
+
+    Attributes:
+        name (str):
+            Name to use when reports about this metric are printed.
+        goal (sdmetrics.goal.Goal):
+            The goal of this metric.
+        min_value (Union[float, tuple[float]]):
+            Minimum value or values that this metric can take.
+        max_value (Union[float, tuple[float]]):
+            Maximum value or values that this metric can take.
+    """
+
+    name: 'CardinalityBoundaryAdherence'
+    goal: Goal.MAXIMIZE
+    min_value: 0.0
+    max_value: 1.0
+
+    @staticmethod
+    def compute_breakdown(real_data, synthetic_data):
+        """Calculate the percentage of synthetic parents with cardinality in the correct range.
+
+        Args:
+            real_data (tuple(pd.Series, pd.Series)):
+                A tuple with the real primary key Series as the first element and real
+                foreign keys Series as the second element.
+            synthetic_data (tuple(pd.Series, pd.Series)):
+                A tuple with the synthetic primary key as the first element and synthetic
+                foreign keys as the second element.
+
+        Returns:
+            Union[float, tuple[float]]:
+                Metric output.
+        """
+        real_cardinality = pd.DataFrame(index=real_data[0].copy())
+        real_cardinality['cardinality'] = real_data[1].value_counts()
+        real_cardinality = real_cardinality.fillna(0)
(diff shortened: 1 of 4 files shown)


sdmetrics/column_pairs/statistical/cardinality_boundary_adherence.py
```diff
@@ -0,0 +1,80 @@
+"""ColumnPair metrics based on Kullback–Leibler Divergence."""
```
R-Palazzo, line 1: Update this line

sdmetrics/column_pairs/statistical/cardinality_boundary_adherence.py
```diff
@@ -0,0 +1,80 @@
...
+        max_value (Union[float, tuple[float]]):
+            Maximum value or values that this metric can take.
+    """
+
+    name: 'CardinalityBoundaryAdherence'
+    goal: Goal.MAXIMIZE
+    min_value: 0.0
+    max_value: 1.0
+
+    @staticmethod
```
R-Palazzo, line 31: @amontanez24 some metrics defines `compute_breakdown` as a `staticmethod` and other as `classmethod`. Should we be consistent? Most of the metrics use `classmethod`.

sdmetrics/column_pairs/statistical/cardinality_boundary_adherence.py
```diff
@@ -0,0 +1,80 @@
...
+        Args:
+            real_data (tuple(pd.Series, pd.Series)):
+                A tuple with the real primary key Series as the first element and real
+                foreign keys Series as the second element.
+            synthetic_data (tuple(pd.Series, pd.Series)):
+                A tuple with the synthetic primary key as the first element and synthetic
+                foreign keys as the second element.
+
+        Returns:
+            Union[float, tuple[float]]:
```
R-Palazzo, line 44: `dict`

sdmetrics/column_pairs/statistical/cardinality_boundary_adherence.py
```diff
@@ -0,0 +1,80 @@
...
```
R-Palazzo, line 77: `float`

R-Palazzo (commented):
Looks good!

sdmetrics/column_pairs/statistical/cardinality_boundary_adherence.py
amontanez24: we can make consistent. If any of them require info from the class then it should be a class method, but if none of them do then I guess it should be a staticmethod

R-Palazzo: Yes at least the `CorrelationSimilarity`: https://github.com/sdv-dev/SDMetrics/blob/b1853d94166c04b8cbd2e22581ac269e9a6db5d8/sdmetrics/column_pairs/statistical/correlation_similarity.py#L79

amontanez24: Interesting, that case I think all the methods involved could be static. It doesn't seem like it ever uses an attribute of the class

frances-h: I think it makes sense on a case by case basis. If the needs to access other methods or properties of the class, it should be a `classmethod` for convenience. Otherwise it can just be a `staticmethod` since we don't need to access anything else on the class.

R-Palazzo (approved):

R-Palazzo (approved):
LGTM! I think this branch should be merged to the `diagnostic_report_updates` branch

sdmetrics/column_pairs/statistical/cardinality_boundary_adherence.py
amontanez24: Yeah, although I think in the case of `CorrelationSimilarity` all the methods involved could be static. None of them seem to be using properties of the class

PR #488, opened 2023-10-31
Add `Relationship Validity` property

### LINKED ISSUE #469: Add `Relationship Validity` property
### Problem Description
_As a user, I'd like a property that measures the validity of the relationships in my synthetic data._

After we have created the CardinalityBoundaryAdherence metric (#464) and the ReferentialIntegrity metric (#461), we would like to use them to measure this property.

### Expected behavior
- Add a new property `Relationship Validity` for multi table reports.

#### Attributes
The property should have the following attributes:
- `_num_iteration_case = 'relationship'`

#### Score Calculation
For every relationship, the property should calculate both the `CardinalityBoundaryAdherence` and the `ReferentialIntegrity` metrics. 

#### Methods
- `_generate_details(self, real_data, synthetic_data, metadata, progress_bar=None)`: This method will generate the `details` dataframe for the property. 
        - `Parent Table`: The parent table name
        - `Child Table`: The child table name
        - `Primary Key`: The primary key column associated with this relationship
        - `Foreign Key`: The foreign key column associated with this relationship 
        - `Metric`: The metric that was calculated
    - The progress bar should update for after each relationship has been evaluated.
- `get_visualization(self, table_name)`: This method should return a bar graph of the scores for each relationship, similar to the `Cardinality` property
    - The `table_name` is required. Show all relationships where `table_name` is either the parent or the child table.  
    - Label the bars based on the relationship. (Align with the `Cardinality` property from the Quality report.)
    - On hover, there should be more details about the metric applied and the score
- `get_details(self, table_name=None)`: If no `table_name` is provided, return all details. If `table_name` is provided, return all rows that have the given table as either the parent or child.

R-Palazzo:
CU-86ayg9y9r
Resolve #469

The visualization is @npatki :
Screenshot 2023-10-31 at 09 18 58

### DIFF
--- a/sdmetrics/reports/multi_table/_properties/base.py
+++ b/sdmetrics/reports/multi_table/_properties/base.py
@@ -54,6 +54,14 @@ class BaseMultiTableProperty():
                 iterations += (len(parent_columns) * len(child_columns))
             return iterations
 
+    @staticmethod
+    def _extract_tuple(data, relation):
+        parent_data = data[relation['parent_table_name']]
+        child_data = data[relation['child_table_name']]
+        return (
+            parent_data[relation['parent_primary_key']], child_data[relation['child_foreign_key']]
+        )
+
     def _compute_average(self):
         """Average the scores for each column."""
         is_dataframe = isinstance(self.details, pd.DataFrame)
@@ -146,22 +154,6 @@ class BaseMultiTableProperty():
             )
 
         return self._properties[table_name].get_visualization()
-
-    def _get_details_for_table_name_with_relationships(self, table_name):
-        """Return the details for the given table name.
-
-        Args:
-            table_name (str):
-                Table name to get the details for.
-
-        Returns:
-            pandas.DataFrame:
-                The details for the given table name.
-        """
-        if all(column in self.details.columns for column in ['Child Table', 'Parent Table']):
-            is_child = self.details['Child Table'] == table_name
-            is_parent = self.details['Parent Table'] == table_name
-            return self.details[is_child | is_parent].copy()
 
     def get_details(self, table_name=None):
         """Return the details table for the property for the given table.
--- a/sdmetrics/reports/multi_table/_properties/base.py
+++ b/sdmetrics/reports/multi_table/_properties/base.py
@@ -147,6 +147,22 @@ class BaseMultiTableProperty():
 
         return self._properties[table_name].get_visualization()
 
+    def _get_details_for_table_name_with_relationships(self, table_name):
+        """Return the details for the given table name.
+
+        Args:
+            table_name (str):
+                Table name to get the details for.
+
+        Returns:
+            pandas.DataFrame:
+                The details for the given table name.
+        """
+        if all(column in self.details.columns for column in ['Child Table', 'Parent Table']):
+            is_child = self.details['Child Table'] == table_name
+            is_parent = self.details['Parent Table'] == table_name
+            return self.details[is_child | is_parent].copy()
+
     def get_details(self, table_name=None):
         """Return the details table for the property for the given table.
 
--- a/sdmetrics/reports/multi_table/_properties/cardinality.py
+++ b/sdmetrics/reports/multi_table/_properties/cardinality.py
@@ -67,22 +67,8 @@ class Cardinality(BaseMultiTableProperty):
             'Error': error_messages,
         })
 
-    def get_details(self, table_name=None):
-        """Return the details for the property.
-
-        Args:
-            table_name (str):
-                Table name to get the details for.
-                Defaults to ``None``.
-
-        Returns:
-            pandas.DataFrame:
-                The details for the property.
-        """
-        if table_name is None:
-            return self.details.copy()
-
-        return self._get_details_for_table_name_with_relationships(table_name)
+        if self.details['Error'].isna().all():
+            self.details = self.details.drop('Error', axis=1)
 
     def _get_table_relationships_plot(self, table_name):
         """Get the table relationships plot from the parent child relationship scores for a table.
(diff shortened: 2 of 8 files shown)


sdmetrics/reports/multi_table/_properties/cardinality.py
```diff
@@ -65,37 +67,8 @@ def _generate_details(self, real_data, synthetic_data, metadata, progress_bar=No
...
-        Returns:
-            pandas.DataFrame:
-                The details for the property.
-        """
-        if table_name is None:
-            return self.details.copy()
-
-        return self._get_details_for_table_name(table_name)
+        if self.details['Error'].isna().all():
+            self.details = self.details.drop('Error', axis=1)
```
frances-h, line 71: Is there a way we can move this to the base class? I think we did it in the single table case, since every property will run this.

R-Palazzo: Good point, it's actually done for multi-table here:
https://github.com/sdv-dev/SDMetrics/blob/5f7488a35fd1ebd71e12d8fcd153e54815eb5839/sdmetrics/reports/multi_table/_properties/base.py#L114
So I just removed the lines in [9718600](https://github.com/sdv-dev/SDMetrics/pull/488/commits/97186002d26008a2eb06d08e85d6741623bf8043)

sdmetrics/reports/multi_table/_properties/base.py
```diff
@@ -45,6 +45,14 @@ def _get_num_iterations(self, metadata):
                 iterations += (len(parent_columns) * len(child_columns))
             return iterations
 
+    @staticmethod
+    def _extract_tuple(data, relation):
+        parent_data = data[relation['parent_table_name']]
+        child_data = data[relation['child_table_name']]
```
amontanez24, line 51: should we error if the parent or child table is not in the data?

tests/integration/reports/multi_table/_properties/test_relationship_validity.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_with_progress_bar(self):
+        """Test that the progress bar is correctly updated."""
+        # Setup
+        real_data, synthetic_data, metadata = load_demo(modality='multi_table')
+        relationship_validity = RelationshipValidity()
+        num_relationship = 2
+
+        progress_bar = tqdm(total=num_relationship)
+        mock_update = Mock()
```
amontanez24, line 31: Instead of mocking I think we can use a fixture like [capsys](https://docs.pytest.org/en/7.4.x/how-to/capture-stdout-stderr.html) and try to evaluate the actual printed output

tests/unit/reports/multi_table/_properties/test_relationship_validity.py
```diff
@@ -0,0 +1,302 @@
...
+            'Primary Key': ['user_id', 'user_id'],
+            'Foreign Key': ['user_id', 'user_id'],
+            'Metric': ['ReferentialIntegrity', 'CardinalityBoundaryAdherence'],
+            'Score': [0.7, 0.3],
+        })
+
+        assert score == 0.5
+        progress_bar.update.assert_called()
+        assert progress_bar.update.call_count == 1
+        mock_compute_average.assert_called_once_with()
```
amontanez24, line 177: minor: if it's empty I think you can do
```
mock_compute_average.assert_called_once()
```

amontanez24 (commented):
Looking really good!

sdmetrics/reports/multi_table/_properties/base.py
R-Palazzo: I feel that this is the report's job to check that the relationships are correct.
We have this method:
https://github.com/sdv-dev/SDMetrics/blob/5f7488a35fd1ebd71e12d8fcd153e54815eb5839/sdmetrics/reports/multi_table/base_multi_table_report.py#L17
That does something a bit different but should error if the parent or child table is not in the data I think.
so when we call `_extract_tuple()` this has already been checked.

tests/integration/reports/multi_table/_properties/test_relationship_validity.py
R-Palazzo: Yes, done in [a5afedd](https://github.com/sdv-dev/SDMetrics/pull/488/commits/a5afeddfa288b600e992ca10294b44f622cec736)

tests/unit/reports/multi_table/_properties/test_relationship_validity.py
R-Palazzo: Yes, done in [a5afedd](https://github.com/sdv-dev/SDMetrics/pull/488/commits/a5afeddfa288b600e992ca10294b44f622cec736)

sdmetrics/reports/multi_table/_properties/base.py
R-Palazzo: BTW I think this method (`_validate_relationships`) is a bit old, maybe we want to update it?

amontanez24 (approved):
LGTM!

sdmetrics/reports/multi_table/_properties/relationship_validity.py
```diff
@@ -0,0 +1,134 @@
...
+                        real_columns,
+                        synthetic_columns,
+                    )
+                    error_message = None
+                except Exception as e:
+                    relation_score = np.nan
+                    error_message = f'{type(e).__name__}: {e}'
+                finally:
+                    if progress_bar is not None and idx % 2 == 0:
+                        progress_bar.update()
```
frances-h, line 58: Instead of `idx %2 == 0` could we do `idx % len(metrics) == 0? Or maybe just remove the update from this for loop entirely? I think rather than running this as a `finally` we can move it outside of the metric `for` loop and just update for each relationship and then also change `for idx, metric in enumerate(metrics)` to just `for metric in metrics`.

tests/unit/reports/multi_table/_properties/test_relationship_validity.py
```diff
@@ -0,0 +1,302 @@
...
+            'Foreign Key': ['user_id', 'user_id'],
+            'Metric': ['ReferentialIntegrity', 'CardinalityBoundaryAdherence'],
+            'Score': [np.nan, np.nan],
+            'Error': ['ValueError: error 1', 'ValueError: error 2']
+        })
+
+        assert pd.isna(score)
+        pd.testing.assert_frame_equal(relationship_validity.details, expected_details_property)
+        progress_bar.update.assert_called()
+        assert progress_bar.update.call_count == 1
```
frances-h, line 221: same

R-Palazzo:
Thanks for the review @frances-h, I addressed the comments ;)

frances-h (approved):

PR #495, opened 2023-11-02
Multi table quality report should handle multi-foreign keys (to same parent)

### LINKED ISSUE #406: Multi table quality report should handle multi-foreign keys (to same parent)
### Problem Description
Currently the [Cardinality property](https://docs.sdv.dev/sdmetrics/reports/quality-report/whats-included#table-relationships) in the multi-table quality report assumes that there is only 1 connection between every parent and child table. This is not always true.

It's possible that a child table has multiple foreign keys that point to the _same primary key column_ in the parent. For example: I can have a parent table `banks` and a child table `transactions`. Then for bank-to-bank transactions, there should be 2 foreign keys in `transactions` that point point to `banks` (they represent the payor and payee).

### Expected behavior
The Quality Report should be updated to account for this case.

In `get_details`, we expect to show a DataFrame for each breakdown. This table should include a `Foreign Key` column to distinguish relationships that have the same parent and child tables. (Note that we can still use `table_name` to select the portions of the dataframe that match either the parent or child table.)

image

In `get_visualization`, each bar is currently labeled with child and parent. We should also update it with the name of the foreign key. Eg. `transactions (payor) -> banks`

image

R-Palazzo:
CU-85ztmg3ty
Resolve #406
Screenshot 2023-11-02 at 15 03 00

### DIFF
--- a/sdmetrics/reports/multi_table/_properties/cardinality.py
+++ b/sdmetrics/reports/multi_table/_properties/cardinality.py
@@ -33,7 +33,7 @@ class Cardinality(BaseMultiTableProperty):
             float:
                 The average score for the property for all the individual metric scores computed.
         """
-        child_tables, parent_tables = [], []
+        child_tables, parent_tables, child_foreign_key = [], [], []
         metric_names, scores, error_messages = [], [], []
         for relation in metadata.get('relationships', []):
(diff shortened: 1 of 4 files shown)


pvk-developer (approved):

fealho (approved):
🙌

PR #496, opened 2023-11-03
Update `DiagnosticReport` to calculate base correctness of synthetic data

### LINKED ISSUE #471: Update `DiagnosticReport` to calculate base correctness of synthetic data
### Problem Description
_As a user, I want a diagnostic report that evaluates the base correctness of my synthetic data._

This report will replace the existing `DiagnosticReport`. It will also be more similar to the `QualityReport`. This report should include the `DataValidity` property (#467), the `Structure` property (#468), and, in the multi-table case, the `RelationshipValidity` property (#469).

### Expected behavior
The report should calculate the following properties:
- Properties:
    - DataValidity
    - Stucture
    - RelationshipValidity (multi-table report only)

In addition, the report API should be changed to more closely match what the QualityReport currently does:
- Results handler: The results handler should print out the overall score for the report, followed by the individual scores for each property similar to what the QualityReport does now. 
    -  The results should now print `'Overall Score:'` instead of `'Overall <report type> Score:'`. This will change for the QualityReport as well.
- The `get_results` method should be renamed to `get_score` and should return the overall score for the report.
- The `get_properties` and `get_details` methods should work the same for both reports

```python
>>> from sdmetrics.reports.multi_table import DiagnosticReport
>>> report.generate(real_data, synthetic_data, metadata)

Generating report ...
(1/3) Evaluating Data Validity: : 100%|██████████| 54/54
(2/3) Evaluating Data Structure: : 100%|██████████| 3/3
(3/3) Evaluating Relationship Validity: : 100%|██████████| 2/2

Overall Score: 100.0%

Properties:
- Data Validity: 100.0%
- Data Structure: 100.0%
- Relationship Validity: 100%
```

R-Palazzo:
CU-86aygcpga
Resolve #471


frances-h (approved):
Looks good!

amontanez24 (approved):
LGTM! Glad we were able to reduce the code by so much

PR #500, opened 2023-11-07
Visualize sequential data with line plot

### LINKED ISSUE #376: Add visualization for timeseries / sequential data
### Problem Description
If my data represents sequences, it's hard to evaluate whether the real and synthetic data are similar. Offering a visualization would immediately help me see what the differences are.

### Expected behavior
When comparing a single sequence of real and synthetic data, a simple line plot would suffice. See [plotly line plots](https://plotly.com/python/line-charts/). The x-axis would be the sequence index (if not present, then an index of `[0, 1, 2 ...]`) and the y-axis wold represent a column of data. We can use a similar color scheme as the other visualizations.

Example (we'd need to add titles for graph and axes)
image

When comparing multiple sequences of real with multiple sequences of synthetic, we can assume the user is providing sequences that are already filtered by some criterion (eg. they are already clustered) so they should be similar. Then we could plot the average of each value, with shading to denote the min/max.

Example (we'd need to add titles for graph and axes)
image

### Proposed API
We could implement a `utils.get_column_line_plot` for this with the following inputs:

- (required) `real_data`: A pandas.DataFrame containing the table of your real data
- (required) `synthetic_data`: A pandas.DataFrame containing the synthetic data
- (required) `column_name`: The name of the column you want to plot. **Only datetime and numerical values can be supported with a line plot**
- (required) `metadata`: A dictionary of Single Table Metadata

With the following logic:
- If the metadata contains a `sequence_index` then that should be used for the x-axis. Otherwise, just use an index of `[0, 1, 2, ...]`.
- If the metadata contains a `sequence_key` then you are comparing multiple sequences of data (as determine by the key). Otherwise, you are comparing single sequences.
- If there are missing values present, we should denote the % for real and synthetic, just like we do for [get_column_plot](https://github.com/sdv-dev/SDMetrics/blob/master/sdmetrics/reports/utils.py#L232)

lajohn4747:
CU-85ztcte5r

resolves #376 

This PR adds the ability to generate a line graph for sequential data separated out by real and synthetic data. Able to graph against a single column and with the metadata, we can defined the x-axis and other sequences within our data. Multiple sequences will cause the graph to graph the average but show a range from min-max of the real/synthetic data.

Screenshot 2023-11-08 at 12 59 58 AM


amontanez24:
Task linked: [CU-85ztcte5r SDMetrics - Add visualization for timeseries / sequential data  #376](https://app.clickup.com/t/85ztcte5r)

sdmetrics/visualization.py
```diff
@@ -478,3 +479,186 @@ def get_column_pair_plot(real_data, synthetic_data, column_names, plot_type=None
...
+
+    # Add min-max shading
+    if 'min' in all_data and 'max' in all_data:
+        fig.add_trace(
+            go.Scatter(
+                name='Real-Min',
+                x=real_data[x_axis],
+                y=real_data['min'],
+                hoverinfo='skip',
+                marker={'color': 'rgba(0, 0, 54, 0.25)'},
```
amontanez24, line 540: should we store these colors as constants?

tests/unit/test_visualization.py
```diff
@@ -700,3 +773,187 @@ def test_get_column_pair_plot_plot_type_is_box(mock__generate_heatmap_plot):
...
+
+    real_data_submitted = real_data.copy()
+    synthetic_data_submitted = synthetic_data.copy()
+    real_data_submitted['sequence_index'] = real_data_submitted.index
+    synthetic_data_submitted['sequence_index'] = synthetic_data_submitted.index
+    real_data_submitted['Data'] = 'Real'
+    synthetic_data_submitted['Data'] = 'Synthetic'
+
+    # Assert
+    # Setup
```
amontanez24, line 950: get rid of setup comment

amontanez24 (approved):
LGTM!

pvk-developer (approved):
LGTM!

PR #519, opened 2023-11-15
Update `TableFormat` metric to `TableStructure` + fix its computation

### LINKED ISSUE #518: Update `TableFormat` metric to `TableStructure` + fix its computation
### Problem Description
This metric is checking the names of the columns as well as the dtypes of the data. The dual-purpose is confusing, especially since (a) it makes the computation more difficult, and (b) there are other metrics that look for the validity of the data.

Let's simplify the metric.

### Expected behavior

1. Rename the metric to `TableStructure`
2. Stop checking for dtypes in the this metric. We only need to check for the column names.
    - We can get rid of the parameter for ignoring dtypes
    - The `'Structure'` property in the Diagnostic Report no longer has to compute or pass the parameter in anymore
 3. Fix the computation to be

```
score = (R intersection S) / (R union S)
``` 

Where _R_ is the set of real column names and _S_ is the set of synthetic column names.

R-Palazzo:
CU-86ayq4d11
Resolve #518


sdv-team:
Task linked: [CU-86ayq4d11 SDMetrics - Update `TableFormat` metric to `TableStructure` + fix its computation #518](https://app.clickup.com/t/86ayq4d11)

frances-h (approved):

sdmetrics/single_table/table_structure.py
```diff
@@ -0,0 +1,60 @@
...
+    goal = Goal.MAXIMIZE
+    min_value = 0
+    max_value = 1
+
+    @classmethod
+    def compute_breakdown(cls, real_data, synthetic_data):
+        """Compute the score breakdown of the table format metric.
+
+        Args:
+        real_data (pandas.DataFrame):
```
fealho, line 33: Extra tab under `Args`

fealho (approved):

PR #527, opened 2023-11-22
Add metadata validation

### LINKED ISSUE #526: Add metadata validation
Add a private method to validate the metadata for single and multi-table reports. 
For single table the metadata must be a dict with a 'column' key.
For multi-table the metadata must be a dict with a 'table' key. All table must also have a 'column' key

R-Palazzo:
CU-86ayrkg1a
Resolve #526

### DIFF
--- a/sdmetrics/reports/multi_table/base_multi_table_report.py
+++ b/sdmetrics/reports/multi_table/base_multi_table_report.py
@@ -32,12 +32,28 @@ class BaseMultiTableReport(BaseReport):
                 return
 
         error_message = (
-            f'Multi table report {self.__class__.__name__} expects real and synthetic data to be'
+            f'Multi table {self.__class__.__name__} expects real and synthetic data to be'
             ' dictionaries of pandas.DataFrame. If your real and synthetic data are pd.DataFrame,'
             f' please use the single-table {self.__class__.__name__} instead.'
         )
 
         raise ValueError(error_message)
+
+    def _validate_metadata_format(self, metadata):
+        """Validate the metadata."""
+        if not isinstance(metadata, dict):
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86ayrkg1a SDMetrics - Add metadata validation #526](https://app.clickup.com/t/86ayrkg1a)

sdmetrics/reports/multi_table/base_multi_table_report.py
```diff
@@ -32,13 +32,29 @@ def _validate_data_format(self, real_data, synthetic_data):
...
+        if 'tables' not in metadata:
+            raise ValueError(
+                'Multi table reports expect metadata to contain a "tables" key with a mapping'
+                ' from table names to metadata for each table.'
+            )
+        for table_name, table_metadata in metadata['tables'].items():
+            if 'columns' not in table_metadata:
+                raise ValueError(
+                    f'The metadata for table "{table_name}" is missing a "columns" key.'
+                )
```
amontanez24, line 56: @npatki should we error in the case where the metadata is malformed? More specifically, f a users gives metadata with an empty table or empty column, should we error?

amontanez24 (approved):
🚢

sdmetrics/reports/multi_table/base_multi_table_report.py
npatki: @amontanez24 yeah that's probably a good idea. Though I'm neutral on whether or not there's an explicit error message for this case vs. the report just naturally crashing because it cannot access those key names.

npatki (approved):
LGTM

PR #537, opened 2024-02-29
Migrate to pyproject.toml

### LINKED ISSUE #534: Transition from using setup.py to pyroject.toml to specify project metadata
### Problem Description
- We currently use **setup.py** and **setup.cfg** to specify the project metadata. This includes requirements which are scatter in `apt.txt`, `requirements.txt`, and `tox.ini`.
- The Python community selected pyproject.toml as a standard way of specifying project metadata, removing the need for all these files. All the project metadata should go in this file. 
- Other modern packaging and dependency management projects, like [Python Poetry](https://python-poetry.org/), only support the more declarative way of using a pyproject.toml file.

### Expected behavior
- Use `pyproject.toml` as the one file for managing project metadata

lajohn4747:
resolves #534 
Installed successfully with `make install-develop`
Ran tests with `make test`


tasks.py
```diff
@@ -2,9 +2,9 @@
 import inspect
 import operator
 import os
-import re
-import pkg_resources
-import platform
+import toml
```
gsheni, line 5: use [tomli](https://pypi.org/project/tomli/), it has been more recently updated
```suggestion
import tomli
```

tasks.py
```diff
@@ -38,51 +38,48 @@ def integration(c):
...
+                (spec.version for spec in req.specifier if spec.operator in ('>=', '==')), existing_version)
+            if new_version > existing_version:
+                # Change when a valid newer version is found
+                min_versions[req.name] = f'{req.name}=={new_version}'
+
+    return list(min_versions.values())
 
 
 @task
 def install_minimum(c):
```
amontanez24, line 73: can we copy the tests we added for installing minimum in sdgym?

.github/workflows/minimum.yml
```diff
@@ -21,6 +21,7 @@ jobs:
     - name: Install dependencies
       run: |
           python -m pip install --upgrade pip
-          python -m pip install invoke .[test]
+          python -m pip install Cython==0.29.24
+          python -m pip install invoke .[test] --no-build-isolation
```
amontanez24, line 25: why do we need cython?

lajohn4747: For scikit-learn 0.24 it needs Cython version less than 3.0.0 (it goes 0.29 -> 3.0.0 for some reason) otherwise it won't compile. I am fixing it in the build-system to make minimum versions pass

amontanez24: That's weird, I don't think we should need to have Cython because when we install scikit-learn it should already be built

pyproject.toml
```diff
@@ -24,8 +24,7 @@ dependencies = [
     "pandas>=1.1.3;python_version<'3.10'",
     "pandas>=1.3.4;python_version>='3.10' and python_version<'3.11'",
     "pandas>=1.5.0;python_version>='3.11'",
-    "scikit-learn>=0.24,<2;python_version<='3.10'",
-    "scikit-learn>=1.1.3,<2;python_version>='3.11'",
+    "scikit-learn>0.24,<2",
```
amontanez24: I think we have to cap for the minimum, otherwise when the script runs on python 3.11 it will try to install 0.24. 

I realize something was wrong with the script before and it was skipping the reinstallation of scikit-learn. I don't think 0.24 is actually available on 3.10 so I think we just have to update the requirements

lajohn4747: I will try to bump up the scikit version. Running into a similar issue to [this](https://github.com/scikit-learn/scikit-learn/issues/26858).

pyproject.toml
```diff
@@ -0,0 +1,112 @@
...
+pomegranate = ['pomegranate>=0.14.1,<0.14.7']
+
+test = [
+    'sdmetrics[torch]',
+    'pytest>=6.2.5,<7',
+    'pytest-cov>=2.6.0,<3',
+    'pytest-rerunfailures>=10',
+    'jupyter>=1.0.0,<2',
+    'rundoc>=0.4.3,<0.5',
+    'tomli>=2.0.0,<3',
```
R-Palazzo, line 62: Add `pytest-runner>=2.11.1` dependency (previously on the build requirement

setup.cfg
```diff
@@ -57,6 +57,4 @@ use_parentheses = True
 [aliases]
 test = pytest
 
-[tool:pytest]
-collect_ignore = ['setup.py']
```
R-Palazzo, line 61: The `[isort]`, `[pydocstyle]` can go to `pyproject.toml` also

tasks.py
```diff
@@ -2,9 +2,10 @@
 import inspect
 import operator
 import os
-import re
-import pkg_resources
-import platform
+import tomli
```
R-Palazzo, line 5: You can sort the import with `isort tasks.py`

pyproject.toml
```diff
@@ -0,0 +1,112 @@
...
+    'plotly>=5.10.0,<6',
+]
+
+[project.urls]
+"Source Code" = "https://github.com/sdv-dev/SDMetrics"
+"Issue Tracker" = "https://github.com/sdv-dev/SDMetrics/issues"
+"Twitter" = "https://twitter.com/sdv_dev"
+"Chat" = "https://bit.ly/sdv-slack-invite"
+
+[build-system]
```
R-Palazzo, line 42: This can go before `[project]` to be consistent with the other libraries

.github/workflows/readme.yml
```diff
@@ -22,5 +22,6 @@ jobs:
       run: |
           python -m pip install --upgrade pip
           python -m pip install invoke rundoc .
+          python -m pip install tomli
```
R-Palazzo, line 25: I also had to add the packaging dependency here I think

pyproject.toml
```diff
@@ -0,0 +1,112 @@
...
+requires = ['setuptools', 'wheel']
+build-backend = "setuptools.build_meta"
+
+[project.optional-dependencies]
+torch = [
+    "torch>=1.8.0;python_version<'3.10'",
+    "torch>=1.11.0;python_version>='3.10' and python_version<'3.11'",
+    "torch>=2.0.0;python_version>='3.11'",
+]
+
```
R-Palazzo, line 52: remove blank line

pyproject.toml
```diff
@@ -0,0 +1,112 @@
...
+pomegranate = ['pomegranate>=0.14.1,<0.14.7']
+
```
R-Palazzo, line 54: remove blank line

pyproject.toml
```diff
@@ -0,0 +1,112 @@
...
+    'sdmetrics[torch]',
+    'pytest>=6.2.5,<7',
+    'pytest-cov>=2.6.0,<3',
+    'pytest-rerunfailures>=10',
+    'jupyter>=1.0.0,<2',
+    'rundoc>=0.4.3,<0.5',
+    'tomli>=2.0.0,<3',
+    'invoke',
+]
+
```
R-Palazzo, line 65: remove blank line

R-Palazzo (commented):
Looks good!

Some comments are just to be consistent with other libraries.

pyproject.toml
```diff
@@ -0,0 +1,123 @@
...
+    "scipy>=1.5.4,<2;python_version<'3.10'",
+    "scipy>=1.9.2,<2;python_version>='3.10'",
+    'copulas>=0.9.0,<0.10',
+    'tqdm>=4.15,<5',
+    'plotly>=5.10.0,<6',
+]
+
+[project.urls]
+"Source Code" = "https://github.com/sdv-dev/SDMetrics"
+"Issue Tracker" = "https://github.com/sdv-dev/SDMetrics/issues"
```
gsheni, line 42:
```suggestion
"Issue Tracker" = "https://github.com/sdv-dev/SDMetrics/issues"
"Changes" = "https://github.com/sdv-dev/SDMetrics/blob/main/HISTORY.md"
```

gsheni (approved):

pyproject.toml
```diff
@@ -0,0 +1,124 @@
...
+requires-python = ">=3.8,<3.12"
+keywords = ['sdmetrics', 'sdmetrics', 'SDMetrics']
+dependencies = [
+    "numpy>=1.20.0,<2;python_version<'3.10'",
+    "numpy>=1.23.3,<2;python_version>='3.10'",
+    "pandas>=1.1.3;python_version<'3.10'",
+    "pandas>=1.3.4;python_version>='3.10' and python_version<'3.11'",
+    "pandas>=1.5.0;python_version>='3.11'",
+    "scikit-learn>=0.24,<2;python_version<'3.10'",
+    "scikit-learn>=1.1.3,<2;python_version>='3.10'",
```
amontanez24, line 32: I think python 3.10 supports scikit-learn 1.1.0 so if we wanted to be as flexible as possible we can do
```suggestion
    "scikit-learn>=0.24,<2;python_version<'3.10'",
    "scikit-learn>=1.1.0;python_version>='3.10' and python_version<'3.11'",
    "scikit-learn>=1.1.3,<2;python_version>='3.11'",
```

amontanez24 (approved):
One comment but after that :shipit:

PR #540, opened 2024-03-07
Print scores after every progress bar

### LINKED ISSUE #538: Improve readability of the report scores when verbosity is on
### Problem Description
When running the Quality or Diagnostic report in verbose mode (default), the progress and scores are shown. They look something like this:

```
Generating report ...
(1/3) Evaluating Data Validity: : 100%|██████████████████████| 52/52 [00:00<00:00, 4764.17it/s]
(2/3) Evaluating Data Structure: : 100%|█████████████████████| 3/3 [00:00<00:00, 1566.40it/s]
(3/3) Evaluating Relationship Validity: : 100%|██████████████████████| 2/2 [00:00<00:00, 770.16it/s]

Overall Score: 94.67%

Properties:
- Data Validity: 84.01%
- Data Structure: 100.0%
- Relationship Validity: 100.0%
```

This is a bit hard to parse:

1. The 100% at the very top refers to the progress (completion) _not_ the overall property score
2. The property scores printed at the bottom actually refer to the progress bar at the top
3. The overall score is printed in the middle of all this and can be easy to miss

### Expected behavior
A more intuitive approach is to (a) stop printing out the progress %, (b) print out each property score after completing it, and (c) end with the overall score.  So the above report would look something like this:

```
Generating report ...
(1/3) Evaluating Data Validity: |██████████████████████| 52/52 [00:00<00:00, 4764.17it/s]
Score: 84.01%

(2/3) Evaluating Data Structure: |█████████████████████| 3/3 [00:00<00:00, 1566.40it/s]
Score: 100.0%

(3/3) Evaluating Relationship Validity: |██████████████████████| 2/2 [00:00<00:00, 770.16it/s]
Score: 100.0%

Overall Score (Average): 94.67%
```

### Additional context
This same format should be used for the quality report and diagnostic report, for both multi- and single-table modalities.

lajohn4747:
resolve #538 

Print scores according to new spec.

Screenshot 2024-03-25 at 11 05 17 AM

### DIFF
--- a/sdmetrics/reports/base_report.py
+++ b/sdmetrics/reports/base_report.py
@@ -172,9 +172,9 @@ class BaseReport():
         for ind, (property_name, property_instance) in enumerate(self._properties.items()):
             if verbose:
                 num_iterations = int(property_instance._get_num_iterations(metadata))
-                progress_bar = tqdm.tqdm(total=num_iterations, file=sys.stdout)
+                progress_bar = tqdm.tqdm(total=num_iterations, file=sys.stdout, bar_format='{desc}|{bar}{r_bar}|')
                 progress_bar.set_description(
-                    f'({ind + 1}/{len(self._properties)}) Evaluating {property_name}: '
+                    f'({ind + 1}/{len(self._properties)}) Evaluating {property_name}'
                 )
 
             score = self._properties[property_name].get_score(
--- a/sdmetrics/reports/base_report.py
+++ b/sdmetrics/reports/base_report.py
@@ -125,15 +125,8 @@ class BaseReport():
         """
         if verbose:
             sys.stdout.write(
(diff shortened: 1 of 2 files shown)


npatki:
> Just awaiting confirmation that the slight change in spec is ok from @npatki ie. printing Column Pair Trends Score:  instead of Score:

FYI as discussed last week, this is fine by me @amontanez24

npatki (changes requested):
There are a few differences [from the original spec written in the issue](https://github.com/sdv-dev/SDMetrics/issues/538)

1. **Blocking** We _do not_ want to show the progress % in the TQDM bar because this is easily confused with the overall score
2. (Not following spec but I'm OK with it) Spec had just the word `Score` rather than `<Property Name> Score`

R-Palazzo (approved):
LGTM!

PR #564, opened 2024-04-15
Fix `entry_points` warning message

fealho:
CU-86b03bn2u.

The `module_name` attribute doesn't exist for the `entry_point` in `_find_addons`, causing the warning message to fail. Update that to use `version` instead.

This PR also updates the  `flake8-pytest-style` and `isort` versions so that `lint` is consistent across python versions.


sdv-team:
Task linked: [CU-86b03bn2u SDMetrics - Update to work with SDV Enteprise](https://app.clickup.com/t/86b03bn2u)

sdmetrics/__init__.py
```diff
@@ -144,7 +144,7 @@ def _find_addons():
         try:
             addon = entry_point.load()
         except Exception:  # pylint: disable=broad-exception-caught
-            msg = f'Failed to load "{entry_point.name}" from "{entry_point.module_name}".'
+            msg = f'Failed to load "{entry_point.name}" from "{entry_point.value}".'
```
fealho, line 147: `module_name` is only an attribute when `entry_point` is imported from `pkg_resources`. Since I changed it to import from `importlib` instead, that attribute doesn't exist anymore.

amontanez24 (approved):

Issue #581, opened 2024-06-05
Allow me to visualize just the real or synthetic data


npatki:
### Problem Description
All of the [SDMetrics visualization utilities](https://docs.sdv.dev/sdmetrics/reports/visualization-utilities) require that I put in both real and synthetic data in order to see the visualization. However, in some cases, I may be in possession of only the real data (or only the synthetic). In these instances, it would still be useful to use the utilities.

### Expected behavior
The API for the utilities can remain as-is. However, we can accept `None` as a valid value for either `real_data` or `synthetic_data`. The logic can change as follows:

- If the `synthetic_data` is none (but `real_data` is present), then the title can update to say "Real data for ...". Continue to use the dark navy color for real data to keep consistency. It is ok to still show the legend.
- Else if the `real_data` is none (but the `synthetic_data` is present), then the title can update to say "Synthetic data for ...". Continue to use the lighter color for synthetic data to keep consistency. It is ok to still show the legend.
- If both `real_data` and `synthetic_data` are None, then no graph can be generated. In this case, throw an Error:

```
Error: No data provided to plot. Please provide either real or synthetic data.
``` 
 
This should apply to all utility visualizations: `get_column_plot`, `get_column_pair_plot`, and `get_cardinality_plot` 

### Additional context
Until we add this feature, here is a quick-and-dirty workaround: Supply the same dataset with both the `real_data` and `synthetic_data` parameters. Since the rendered plot is interactive, you can then click the legend to "hide" one of the options.

```python
from sdmetrics.visualization import get_column_pair_plot

fig = get_column_pair_plot(
    real_data=real_data,
    synthetic_data=real_data,
    column_names=['column_a', 'column_b'],
    plot_type='scatter'
    
)

fig.show()
```

PR #595, opened 2024-06-20
Allow get_column_pair_plot to visualize one dataset instead of both real and synthetic

lajohn4747:
PR for 1 of 3 graphs to be able show real and synthetic separately using a None for one of the datasets.

This PR is for `get_column_pair_plot` and the relevant tests

Part of issue #581 
CU-86b0t56r9

get_column_pair_plot_heat_both
get_column_pair_plot_heat_real
get_column_pair_plot_heat_synth
get_column_pair_plot_scatter_both
get_column_pair_plot_scatter_real
get_column_pair_plot_scatter_synth


sdv-team:
Task linked: [CU-86b0t56r9 SDMetrics - Allow me to visualize just the real or synthetic data #581](https://app.clickup.com/t/86b0t56r9)

pvk-developer (approved):

sdmetrics/visualization.py
```diff
@@ -86,12 +86,20 @@ def _generate_heatmap_plot(all_data, columns):
...
+    unique_values = all_data['Data'].unique()
+
     fig = px.density_heatmap(
         all_data, x=columns[0], y=columns[1], facet_col='Data', histnorm='probability'
     )
 
+    title = ''
+    for name in unique_values:
+        title += f'{name} vs. '
+    title = title[:-4]
```
amontanez24, line 98:
```suggestion
    title = ' vs. '.join(unique_values)
```

sdmetrics/visualization.py
```diff
@@ -159,8 +168,14 @@ def _generate_scatter_plot(all_data, columns):
         symbol='Data',
     )
 
+    title = ''
+    for name in unique_values:
+        title += f'{name} vs. '
+    title = title[:-4]
```
amontanez24, line 174: same

amontanez24 (commented):
I think we should add tests for only one column for both `_generate_scatter_plot` and `_generate_heatmap_plot`

sdmetrics/visualization.py
```diff
@@ -88,15 +88,15 @@ def _generate_heatmap_plot(all_data, columns):
     """
     unique_values = all_data['Data'].unique()
 
+    if len(columns) != 2:
+        raise ValueError('Generating a heatmap plot requires only two columns for the axis.')
```
amontanez24, line 92: I got confused with this and the other PR about getting a column plot 😅 . I don't think you need to assert that exactly 2 columns are passed and you can get rid of those tests. If you want to leave it can you change the following
```suggestion
    if len(columns) != 2:
        raise ValueError('Generating a heatmap plot requires exactly two columns for the axis.')
```

PR #596, opened 2024-06-20
Allow `get_column_plot` to graph synthetic and real data individually

lajohn4747:
Part 2 of 3 for allowing visualizations to work with a single dataset: #581 
CU-86b0t56r9

This PR is to handle `get_column_plot`

get_column_plot_bar_both
get_column_plot_bar_real
get_column_plot_bar_synth
get_column_plot_dist_both
get_column_plot_dist_real
get_column_plot_dist_synth


sdv-team:
Task linked: [CU-86b0t56r9 SDMetrics - Allow me to visualize just the real or synthetic data #581](https://app.clickup.com/t/86b0t56r9)

npatki:
Hi @lajohn4747 the request was to always have the correct color -- so when you're doing synthetic data, it should be light blue, not dark blue. However, all of your graphs are the darker color. Could you update this?

frances-h (approved):

amontanez24 (commented):
Looks good but it doesn't seem like there are any unit tests for `_generate_column_plot` and that is mocked in the test for `get_column_plot` so a lot of this functionality seems to be untested. This seems to come from before this PR but can you either
1. Add tests for `_generate_column_plot` or
2. Remove the mock and make sure everything is behaving as expected
If you do the second one you can mock px and ff so you don't have to graph anything and can just ealuate the right inputs are being passed

tests/unit/test_visualization.py
```diff
@@ -276,6 +280,191 @@ def test_get_column_plot_no_data():
...
+    ['Real', 'Synthetic'] == mock_distplot.call_args[0][1]
+
+    expected_colors = [PlotConfig.DATACEBO_DARK, PlotConfig.DATACEBO_GREEN]
+    expected_parameters = {
+        'show_hist': False,
+        'show_rug': False,
+        'colors': expected_colors,
+    }
+    assert expected_parameters == mock_distplot.call_args[1]
+    mock_distplot.assert_called_once()
```
amontanez24, line 336: minor: I think for these you can do 
```python
mock_distplot.assert_called_once_with(show_hist=False, show_rug=False ...)
```
The only time I think you'd need to pull out the call args itself is to evaluate dataframes

lajohn4747: Updated so I can use `assert_called_once_with`. I had to update it to use `ANY` from `unittest.mocks`

amontanez24 (approved):
:shipit:

PR #598, opened 2024-06-21
Update Referential Integrity metric to support NaNs in child column 

### LINKED ISSUE #587: Update Referential Integrity metric to support NaNs in child column
### Problem Description
- For the Referential Integrity metric, we currently do the following:
  - If the foreign key in the real data has any NaN values, those are dropped in the foreign key in the synthetic data.
  - We also verify that all foreign keys have valid parent.
- To support NaNs values in foreign keys, we need to update this logic. 

### Expected behavior
- Update `ReferentialIntegrity.compute_breakdown()` to do the following:
  - If the real data has NaN values, and the synthetic data has NaN values in the FK, then it is okay.
  - If the real data does not have NaN values, but the synthetic data has NaN values in the FK, this will result in a lower score. 
    - The score can be lowered by the number of extra NaNs present in the synthetic data.
- Update test_referential_interity.py
- Update documentation

### Additional context
- Function to update:
https://github.com/sdv-dev/SDMetrics/blob/51f3ac2904c065462097c80f4387b973657ad6d9/sdmetrics/column_pairs/statistical/referential_integrity.py#L36

R-Palazzo:
CU-86b0uyzmn
Resolve #587

### DIFF
--- a/tests/unit/column_pairs/statistical/test_referential_integrity.py
+++ b/tests/unit/column_pairs/statistical/test_referential_integrity.py
@@ -120,4 +120,4 @@ class TestReferentialIntegrity:
         # Assert
         assert result_0 == 1.0
         assert result_1 == 0.8
-        assert result_2 == 0.6666666666666666
+        assert result_2 == 2 / 3
--- a/tests/unit/column_pairs/statistical/test_referential_integrity.py
+++ b/tests/unit/column_pairs/statistical/test_referential_integrity.py
@@ -68,32 +68,56 @@ class TestReferentialIntegrity:
         assert result == 0.6
 
     def test_compute_with_nan_foreign_keys_real_data(self):
-        """Test the ``compute`` method with NaN foreign keys inside the real data."""
+        """Test the ``compute`` method with NaN foreign keys inside the real data.
+
+        Here the score should be 1.0 regardless of if the synthetic data has NaN values or not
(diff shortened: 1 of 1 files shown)


sdv-team:
Task linked: [CU-86b0uyzmn SDMetrics - Update ReferentialIntegrity to support NaNs in child column  #587](https://app.clickup.com/t/86b0uyzmn)

tests/unit/column_pairs/statistical/test_referential_integrity.py
```diff
@@ -68,32 +68,56 @@ def test_compute(self, compute_breakdown_mock):
...
+        )
+        result_2 = metric.compute(
+            real_data=(parent_keys, real_fk), synthetic_data=(parent_keys, synth_fk_2_nan)
         )
 
         # Assert
-        assert result == 0.8
+        assert result_0 == 1.0
+        assert result_1 == 0.8
+        assert result_2 == 0.6666666666666666
```
gsheni, line 123:
```suggestion
        assert result_2 == (2/3)
```

rwedge (approved):
Looks good.

rwedge (commented):
~~Can we look at the case where the real data has a NaN in the foreign kyes but the synth does not?~~ Already present

gsheni (approved):

PR #601, opened 2024-06-26
Fix tests failing due to scipy 1.14

lajohn4747:
Failing tests due to scipy 1.14

Tests were failing due to two reasons:
- Scipy changed there error messages
- Scipy `ks_2samp` does not throw value error if nans are passed in anymore
- It seems float point accuracy increased


tests/integration/reports/single_table/test_quality_report.py
```diff
@@ -250,6 +250,12 @@ def test_report_end_to_end_with_errors(self):
 
         real_data['second_perc'].iloc[2] = 'a'
 
+        def get_error_type(error):
```
amontanez24, line 253: since you use this twice should you just pull it out?

amontanez24 (approved):

R-Palazzo (approved):

PR #611, opened 2024-07-24
Check that every property score are index-free

### LINKED ISSUE #583: Check that every property score are index-free
### Problem Description

A property score, like a metric, should not depend on the index of the real and synthetic data. It should only depend on their column values. It would be nice to have some integration tests that check that this is the case for all our current properties (single and multi-table) as well as for future ones. 

### Expected behavior

Write a simple integration test where real and synthetic data are the same and only differ by their indexes. Check that the property score is 1.0. Apply this test to all the properties (keeping in mind that it should work for future ones also). In the case of some property scores not being 1.0:
- Investigate if an easy fix could solve the issue
- Otherwise, an easy fix would be to copy the data and reset the index at the beginning of any reports.

R-Palazzo:
CU-86b0tvz6j
Resolve #583

### DIFF
--- a/tests/integration/test_property.py
+++ b/tests/integration/test_property.py
@@ -38,7 +38,7 @@ def test_shuffling_data_single_table(property):
 
 
 @pytest.mark.parametrize('property', MULTI_TABLE_PROPERTIES)
-def test_property_multi_table(property):
+def test_shuffling_data_multi_table(property):
     """Test the property score is the same when shuffling the data for multi-table."""
     # Setup
     property_instance = property()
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b0tvz6j SDMetrics - Check that every property score are index-free #583](https://app.clickup.com/t/86b0tvz6j)

lajohn4747 (approved):
LGTM, glad it was fixed previously and we only needed a test for this 👍

rwedge (approved):

PR #621, opened 2024-08-09
X-axis for the bar plot should be labeled `Value` instead of `Category`

### LINKED ISSUE #620: X-axis for the bar plot should be labeled `Value` instead of `Category`
### Environment Details
* SDMetrics version: 0.15.0 (latest)

### Error Description
When visualizing a single column's distribution (univariate), users can supply the `plot_type` to switch between rendering as a distplot or as a bar plot. Continuous values (numerical/discrete) are usually defaulted as distplots but sometimes, it is better to show them as bar plots due to the nature of the data -- eg. if it has many outliers or cannot be shown smoothly.

However, the x-axis title on all bar plots is `Category`. This only makes sense for discrete (categorical) columns, not for continuous columns that I'm choosing to visualize as a bar plot.

**Expected Fix**: Change the x-axis title to `Value` instead. This word works for both discrete and continuous columns.

### Steps to reproduce

```python
from sdv.evaluation.single_table import get_column_plot

    modality='single_table',
    dataset_name='fake_hotel_guests'
)

fig = get_column_plot(
    real_data=data,
    synthetic_data=None,
    metadata=metadata,
    column_name='amenities_fee',
    plot_type='bar'
)
    
fig.show()
```
image

R-Palazzo:
CU-86b1kp24w
Resolve #620

### DIFF
--- a/sdmetrics/visualization.py
+++ b/sdmetrics/visualization.py
@@ -254,7 +254,7 @@ def _generate_column_plot(
         plot_title (str, optional):
             Title to use for the plot. Defaults to 'Real vs. Synthetic Data for column {column}'
         x_label (str, optional):
-            Label to use for x-axis. Defaults to 'Category'.
+            Label to use for x-axis. Defaults to 'Value'.
 
     Returns:
         plotly.graph_objects._figure.Figure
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b1kp24w SDMetrics - X-axis for the bar plot should be labeled `Value` instead of `Category` #620](https://app.clickup.com/t/86b1kp24w)

fealho (approved):

frances-h (approved):

PR #625, opened 2024-08-20
Try to improve performance of contingency_similarity

### LINKED ISSUE #622: Try to improve performance of contingency_similarity
### Problem Description

_As a user, I'd like to get the results of my metrics reports as quickly as possible._

We performed an audit on the `QualityReport` since it seemed to be slow. The conclusion was that most of the time is lost in the contingency similarity metric. More specifically, these lines
https://github.com/sdv-dev/SDMetrics/blob/685731fbfe8ae2744793f7ee93b1dd9700a2f0ef/sdmetrics/column_pairs/statistical/contingency_similarity.py#L45-L54

This is the performance report's visualization
image

### Expected behavior

**Without changing the algorithm at all** the goal of this issue is to improve the performance of `contingency_similarity`. Optimizations that are in scope include
- Trying different pandas or numpy functions instead of crosstab
- Trying to do any type conversions at a higher level (eg. the astype(str) calls are happening multiple times on the same columns)
- Seeing if there is a more efficient way to compute the table

The optimizations **should not change** the overall algorithm of the metric.

### Additional context

- If not many optimizations can be made, we can follow up with a different issue

amontanez24:
resolves #622 

Below is the profiling of the `QualityReport` before the changes. As you can see most time was spent in the `crosstab` method. 
old_code_100_000

To improve this, I switched to using `groupby` instead. `Groupby` seems to be faster than `crosstab` and `pivot_table` according to this [post](https://ramiro.org/notebook/pandas-crosstab-groupby-pivot/) and trials I ran. The one issue is that `groupby` returns data as a multi-indexed series instead of a table with the different column values on one axis and the different row values on the other. To compute the metric, I could either use `unstack` to convert the return value of `groupby` to the table we wanted, or change the logic to sum over the multi-indexed series. I tried both examples and show their results below. I used the same dataset for each method.

### Results for all methods at 1000 rows
Screenshot 2024-08-20 at 11 46 11 AM

### Results for all methods at 10,000 rows
Screenshot 2024-08-20 at 11 48 15 AM

### Results for all methods at 1,000,000 rows
image

As you can see, as the number of rows increases, the gap in time between methods decreases. Despite that, groupby was always faster. The specific method of groupby that was fastest seemed to vary based on number of rows.

Below is the profiling for each method.
### Using groupby, unstack and reindexing after unstacking
groupby_unstack_reindex

### Using groupby, unsstack and reindexing before unstacking
groupby_reindex_unstack

### Using groupby on its own
Screenshot 2024-08-20 at 11 57 15 AM

### DIFF
--- a/sdmetrics/column_pairs/statistical/contingency_similarity.py
+++ b/sdmetrics/column_pairs/statistical/contingency_similarity.py
@@ -26,30 +26,91 @@ class ContingencySimilarity(ColumnPairsMetric):
     max_value = 1.0
 
     @classmethod
-    def new_compute(cls, real_data, synthetic_data):
-        columns = real_data.columns[:2]
-        real = real_data[columns]
-        synthetic = synthetic_data[columns]
-        contingency_real = real.groupby(list(columns), dropna=False).size() / len(real)
-        contingency_synthetic = synthetic.groupby(list(columns), dropna=False).size() / len(synthetic)
-        combined_index = contingency_real.index.union(contingency_synthetic.index)
-        contingency_synthetic = contingency_synthetic.reindex(combined_index, fill_value=0)
-        contingency_real = contingency_real.reindex(combined_index, fill_value=0)
-        diff = abs(contingency_real - contingency_synthetic).fillna(0)
-        variation = diff / 2
-        return 1 - variation.sum()
-        # contingency_real = contingency_real.unstack(level=1, fill_value=0)
-        # contingency_synthetic = contingency_synthetic.unstack(level=1, fill_value=0)
-        # all_columns = contingency_real.columns.union(contingency_synthetic.columns)
-        # all_indices = contingency_real.index.union(contingency_synthetic.index)
-        # contingency_real = contingency_real.reindex(
-        #     index=all_indices, columns=all_columns, fill_value=0
-        # )
-        # contingency_synthetic = contingency_synthetic.reindex(
-        #     index=all_indices, columns=all_columns, fill_value=0
-        # )
-        # variation = abs(contingency_real - contingency_synthetic) / 2
-        # return 1 - variation.sum().sum()
+    def _compute_by_method(cls, real_data, synthetic_data, method='old'):
+        if method == 'old':
+            columns = real_data.columns[:2]
+            real = real_data[columns]
+            synthetic = synthetic_data[columns]
+            contingency_real = pd.crosstab(
+                index=real[columns[0]].astype(str),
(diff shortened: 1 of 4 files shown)


amontanez24:
@rwedge @pvk-developer I included all the variants I tried. Ultimately we'll just pick one before merging, but I wanted you to see them and give any feedback. Personally I like the option that only uses groupby and not unstack since it seems simplest.

amontanez24:
here are the .prof files if you want to examine yourself
[Archive.zip](https://github.com/user-attachments/files/16700741/Archive.zip)

sdmetrics/column_pairs/statistical/contingency_similarity.py
```diff
@@ -25,6 +25,93 @@ class ContingencySimilarity(ColumnPairsMetric):
...
+            real = real_data[columns]
+            synthetic = synthetic_data[columns]
+            contingency_real = real.groupby(list(columns), dropna=False).size() / len(real)
+            contingency_synthetic = synthetic.groupby(list(columns), dropna=False).size() / len(synthetic)
+            combined_index = contingency_real.index.union(contingency_synthetic.index)
+            contingency_synthetic = contingency_synthetic.reindex(combined_index, fill_value=0)
+            contingency_real = contingency_real.reindex(combined_index, fill_value=0)
+            diff = abs(contingency_real - contingency_synthetic).fillna(0)
+            variation = diff / 2
+            return 1 - variation.sum()
```
pvk-developer, line 113: I vote for this approach.

rwedge: This seems reasonable to me as well

tests/integration/reports/multi_table/test_quality_report.py
```diff
@@ -272,9 +272,9 @@ def test_quality_report_with_object_datetimes():
     # Assert
     expected_properties = pd.DataFrame({
         'Property': ['Column Shapes', 'Column Pair Trends', 'Cardinality', 'Intertable Trends'],
-        'Score': [0.7978174603174604, 0.45654629583521095, 0.95, 0.4416666666666667],
+        'Score': [0.7978174603174604, 0.45654629583521095, 0.95, 0.4416666666666666],
     })
-    assert score == 0.6615076057048345
+    assert score == 0.6615076057048344
```
amontanez24, line 277: The new method seems to round the last digit down instead of up

pvk-developer (approved):

rwedge (approved):

PR #633, opened 2024-09-20
Add dtype comparison in `TableStructure` metric (used in Diagnostic report)

### LINKED ISSUE #631: Add dtype comparison in `TableStructure` metric (used in Diagnostic report)
### Problem Description
The Diagnostic Report is meant to uncover if there are any basic problems with the synthetic data (eg. that break the guarantees of SDV). One of these basic checks should be to ensure that each column of the synthetic data is stored in the same format as the real data, in terms of dtypes. Because if this is not the same, then it make break downstream usage of synthetic data.

Unfortunately, this is not being checked anywhere in the Diagnostic Report.

### Expected behavior
The [TableStructure](https://docs.sdv.dev/sdmetrics/metrics/metrics-glossary/tablestructure) metric would be the correct place to do the dtype check. To make this an accurate metric, we should change it to the following:

- In the numerator, only consider a column as overlapping (between real and synthetic data) if it has the same name _and_ the same dtype
- The denominator should be all unique combinations of `(column_name, dtype)` that are in the real and synthetic data.

### Additional context
In #518, we seemed to have removed dtypes because we assumed another metric would have this check. This is not the case. Also we said that checking dtypes may cause computational issues, which also does not appear to be the case.

R-Palazzo:
Resolve #631

### DIFF
--- a/sdmetrics/single_table/table_structure.py
+++ b/sdmetrics/single_table/table_structure.py
@@ -36,11 +36,12 @@ class TableStructure(SingleTableMetric):
             synthetic_data (pandas.DataFrame):
                 The synthetic data.
         """
-        synthetic_columns = set(synthetic_data.columns)
-        real_columns = set(real_data.columns)
-        intersection_columns = real_columns & synthetic_columns
-        union_columns = real_columns | synthetic_columns
-        score = len(intersection_columns) / len(union_columns)
+        real_columns_dtypes = set(zip(real_data.columns, map(str, real_data.dtypes)))
+        synthetic_columns_dtypes = set(zip(synthetic_data.columns, map(str, synthetic_data.dtypes)))
+
+        intersection = real_columns_dtypes & synthetic_columns_dtypes
+        union = real_columns_dtypes | synthetic_columns_dtypes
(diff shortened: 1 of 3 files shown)


frances-h (approved):

sdmetrics/single_table/table_structure.py
```diff
@@ -36,11 +36,12 @@ def compute_breakdown(cls, real_data, synthetic_data):
...
-        real_columns = set(real_data.columns)
-        intersection_columns = real_columns & synthetic_columns
-        union_columns = real_columns | synthetic_columns
-        score = len(intersection_columns) / len(union_columns)
+        real_columns_dtypes = set(zip(real_data.columns, map(str, real_data.dtypes)))
+        synthetic_columns_dtypes = set(zip(synthetic_data.columns, map(str, synthetic_data.dtypes)))
+
+        intersection = real_columns_dtypes & synthetic_columns_dtypes
+        union = real_columns_dtypes | synthetic_columns_dtypes
+        score = len(intersection) / len(union)
```
fealho, line 44: Can this be 0/0?

fealho (approved):

PR #643, opened 2024-10-23
 Add metric for sequence length similarity 

### LINKED ISSUE #638: Add metric for sequence length similarity
### Problem Description

### Expected behavior
Add a new metric called `SequenceLengthSimilarity` to SDMetrics. 

**Data compatibility**: ID columns (representing the sequence key)

**Parameters**:
- (required) `real_data`: A column (pd.Series) containing the sequence key of the real data
- (required) `synthetic_data`: A column (pd.Series) containing the sequence key of the synthetic data

**Output**: A score in range `[0, 1]` -- 0 being the worst and 1 being the best

```python
from sdmetrics.single_column import SequenceLengthSimilarity

score = SequenceLengthSimilarity.compute(
  real_data=real_table['patient_id'],
  synthetic_data=synthetic_table['patient_id']
)
```

**How does it work?** The length of a sequence is determined by the number of times the same sequence key occurs. For example if `id_09231` appeared 150 times in the sequence key, then the sequence is of length 150. This metric compares the lengths of all sequence keys in the real data vs. the synthetic data:

1. Calculate the length of each sequence in the real data (call this distribution D_r)
2. Calculate the length of each sequence in the synthetic data (call this distribution D_s)

fealho:
CU-86b2evqnr, Resolve #638.

### DIFF
--- a/tests/unit/timeseries/test_sequence_length_similarity.py
+++ b/tests/unit/timeseries/test_sequence_length_similarity.py
@@ -8,7 +8,7 @@ class TestSequenceLengthSimilarity:
         """Test it returns 1 when real and synthetic data have the same distribution."""
         # Setup
         real_data = pd.Series(['id1', 'id1', 'id2', 'id2', 'id2', 'id3'])
-        synthetic_data = pd.Series(['id1', 'id1', 'id2', 'id3', 'id3', 'id3'])
+        synthetic_data = pd.Series(['id4', 'id4', 'id5', 'id6', 'id6', 'id6'])
 
         # Run
         score = SequenceLengthSimilarity.compute(real_data, synthetic_data)
@@ -19,11 +19,11 @@ class TestSequenceLengthSimilarity:
     def test_compute_low_score(self):
         """Test it for distinct distributions."""
         # Setup
-        real_data = pd.Series(['id1', 'id1', 'id2'])
-        synthetic_data = pd.Series(['id1', 'id2', 'id3'])
+        real_data = pd.Series([f'id{i}' for i in range(100)])
+        synthetic_data = pd.Series(['id100'] * 100)
 
         # Run
         score = SequenceLengthSimilarity.compute(real_data, synthetic_data)
 
         # Assert
-        assert score == 0.5
+        assert score == 0
--- a/tests/unit/timeseries/test_sequence_length_similarity.py
+++ b/tests/unit/timeseries/test_sequence_length_similarity.py
@@ -4,6 +4,18 @@ from sdmetrics.timeseries.sequence_length_similarity import SequenceLengthSimila
 
 
 class TestSequenceLengthSimilarity:
+    def test_compute(self):
+        """Test it runs."""
+        # Setup
+        real_data = pd.Series(['id1', 'id2', 'id2', 'id3'])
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b2evqnr SDMetrics - Add metric for sequence length similarity #638](https://app.clickup.com/t/86b2evqnr)

R-Palazzo (commented):
Looks good!
I just want to check if we should use actual or normalized counts.
Currently, we use non-normalized counts, so the metric score depends on the size of the real and synthetic data; for instance:
```python
real_data = pd.Series(['id1', 'id1', 'id2', 'id2', 'id2', 'id3'] * 2)
synthetic_data = pd.Series(['id4', 'id4', 'id5', 'id6', 'id6', 'id6'])

score = SequenceLengthSimilarity.compute(real_data, synthetic_data)

print(score)
```
We will print `0.33`, while in both real and synthetic data, we have a sequence that is half; one is one-third, and the last is one-sixth of the data size. Let me know if it makes sense. @Neha, do you have any thoughts on it?

npatki:
>  the metric score depends on the size of the real and synthetic data

As it should! The metric is called **SequenceLength**Similarity, so of course the length of the sequences matter.

In this example, `score1` should not be equal to `score2`:
```python
real_data1 = pd.Series(['id1', 'id1', 'id2', 'id2', 'id2', 'id3'])
real_data2 = pd.Series(['id1', 'id1', 'id2', 'id2', 'id2', 'id3']*2)
synthetic_data = pd.Series(['id4', 'id4', 'id5', 'id6', 'id6', 'id6'])

score1 = SequenceLengthSimilarity.compute(real_data1, synthetic_data)
score2 = SequenceLengthSimilarity.compute(real_data2, synthetic_data)
```

This is because the `SequenceLengthSimilarity` metric is computing the _length of each sequence_ and then using `KSComplement` on the resulting distributions.
- Distribution for `real_data1` is: [2, 3, 1] (these are the sequence lengths of `real_data1`)
- Distribution for `real_data2` is: [4, 6, 2]  (these are the sequence lengths of `real_data2`)
- Distribution for `synthetic_data` is: [2, 1, 3] (these are the sequence lengths of `synthetic_data`)

As a result, `score1` should be 1.0 while `score2` should be 0.33333. Normalizing would defeat the purpose of comparing lengths.

As a general note: Based on the Qs being asked yesterday and today, I am really worried about lack of understanding of what this metric is supposed to be doing. I would encourage re-reading the issue to clear up confusion @fealho and @R-Palazzo .

@amontanez24 could you also review if you get the chance?

R-Palazzo (approved):

tests/unit/timeseries/test_sequence_length_similarity.py
```diff
@@ -0,0 +1,29 @@
+import pandas as pd
+
+from sdmetrics.timeseries.sequence_length_similarity import SequenceLengthSimilarity
+
+
+class TestSequenceLengthSimilarity:
+    def test_compute_one(self):
```
amontanez24, line 7: Maybe we should add a test for a score that isn't 1 or 0

amontanez24 (approved):
This lgtm!

PR #647, opened 2024-10-28
Add metric for inter-row MSAS

### LINKED ISSUE #640: Add metric for inter-row MSAS
### Problem Description

### Expected behavior
Add a metric called InterRowMSAS that performs the MSAS algorithm for inter-row differences in a sequence.

**Parameters**:
- `n_rows_diff`: An integer representing the number of rows to consider when taking the difference
    - (default) 1: Take the difference of a row and the one right before it
    - Int > 0: Take the difference between a row `n` and `n + n_rows_diff`
- `apply_log`: Whether to apply a natural log before taking the difference
   - (default) `False`: Do not apply a log. This results in the absolute difference, useful when you expect the data to grow or shrink linearly
   - `True`: Apply a lot before taking the difference. This is recommended when you expect the data to grow or shrink exponentially

```python
from sdmetrics.column_pairs import InterRowMSAS

score = InterRowMSAS.compute(
  n_rows_diff=100,
  apply_log=False
)
```

1. Calculate the difference between row `r` and row `r+x` for each row in the real data. Then take the average over each sequence to form a distribution D_r
2. Do the same for the synthetic data to form a new distribution D_s

fealho:
CU-86b2evvby, Resolve #640.


sdv-team:
Task linked: [CU-86b2evvby SDMetrics - Add metric for inter-row MSAS #640](https://app.clickup.com/t/86b2evvby)

sdmetrics/timeseries/inter_row.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if apply_log:
+            real_values = np.log(real_values)
+            synthetic_values = np.log(synthetic_values)
+
+        def calculate_differences(keys, values):
+            differences = []
+            for key in keys.unique():
+                group_values = values[keys == key].to_numpy()
+                if len(group_values) > n_rows_diff:
```
fealho, line 70: If `n_rows_diff` is larger than the sequence lengths, `np.mean(diff)` is np.nan. The KSComplement doesn't support np.nan values for python 3.8.

fealho, line 70: If n_rows_diff is larger than the sequence lengths, np.mean(diff) is np.nan. The KSComplement doesn't support np.nan values for python 3.8.

amontanez24: Should we raise an error here?

sdmetrics/timeseries/inter_row.py
```diff
@@ -0,0 +1,78 @@
...
+        real_keys, real_values = real_data
+        synthetic_keys, synthetic_values = synthetic_data
+
+        if apply_log:
+            real_values = np.log(real_values)
+            synthetic_values = np.log(synthetic_values)
+
+        def calculate_differences(keys, values):
+            differences = []
+            for key in keys.unique():
```
amontanez24, line 68: Same question as [this one](https://github.com/sdv-dev/SDMetrics/pull/649#discussion_r1823355587) about using groupby and an aggregate mean instead of looping

tests/unit/timeseries/test_inter_row.py
```diff
@@ -0,0 +1,73 @@
+import pandas as pd
+
+from sdmetrics.timeseries.inter_row import InterRowMSAS
+
+
+class TestInterRowMSAS:
+    def test_compute_identical_sequences(self):
```
amontanez24, line 7: can we add a test where the score isn't exactly 0 or one?

sdmetrics/timeseries/inter_row.py
```diff
@@ -0,0 +1,78 @@
...
+                An integer representing the number of rows to consider when taking the difference.
+            apply_log (bool):
+                Whether to apply a natural log before taking the difference.
+
+        Returns:
+            float:
+                The similarity score between the real and synthetic data distributions.
+        """
+        real_keys, real_values = real_data
+        synthetic_keys, synthetic_values = synthetic_data
```
frances-h, line 60: Same as [this comment](https://github.com/sdv-dev/SDMetrics/pull/649/files#r1823349386), should we check the inputs before trying to unpack?

sdmetrics/timeseries/inter_row.py
frances-h: Can we add a test where `n_rows_diff` is larger than a sequence?

sdmetrics/single_column/statistical/kscomplement.py
```diff
@@ -56,7 +57,13 @@ def compute(real_data, synthetic_data):
             real_data = pd.to_numeric(real_data)
             synthetic_data = pd.to_numeric(synthetic_data)
 
-        statistic, _ = ks_2samp(real_data, synthetic_data)
+        try:
+            statistic, _ = ks_2samp(real_data, synthetic_data)
+        except ValueError as e:
+            if str(e) == 'Data passed to ks_2samp must not be empty':
```
frances-h, line 63: Should we check the inputs instead of trying to catch the error from the error message?

fealho: Only old versions of scipy raise this error, newer ones support empty data.

frances-h (approved):
LGTM! Thanks for addressing

R-Palazzo (approved):

PR #649, opened 2024-10-30
Add metric for general MSAS statistics

### LINKED ISSUE #639: Add metric for general MSAS statistics
### Problem Description

### Expected behavior
Add a metric called `StatisticMSAS` that performs the MSAS algorithm for a given statistic.

**Parameters**:
- `statistic`: A string representing the statistic function to use when computing MSAS
    - (default) `'mean'`: The arithmetic mean
    - `'median'`: The median value
    - `'std'`: The standard deviation
    - `'min'`: The min value
    - `'max'`: The max value

```python
from sdmetrics.column_pairs import StatisticMSAS

score = StatisticMSAS.compute(
  statistic='std'
)
```

1. Calculate the statistic value of each sequence in the real data (call this distribution D_r)
2. Calculate the statistic value of each sequence in the synthetic data (call this distribution D_s)

fealho:
CU-86b2evqnt, Resolve #639.

### DIFF
--- a/sdmetrics/timeseries/statistic_msas.py
+++ b/sdmetrics/timeseries/statistic_msas.py
@@ -74,16 +74,21 @@ class StatisticMSAS:
                 f' Choose from [{", ".join(statistic_functions.keys())}].'
             )
 
+        for data in [real_data, synthetic_data]:
+            if (
+                not isinstance(data, tuple)
+                or len(data) != 2
+                or (not (isinstance(data[0], pd.Series) and isinstance(data[1], pd.Series)))
+            ):
+                raise ValueError('The data must be a tuple of two pandas series.')
+
         real_keys, real_values = real_data
         synthetic_keys, synthetic_values = synthetic_data
         stat_func = statistic_functions[statistic]
 
         def calculate_statistics(keys, values):
-            statistics = []
-            for key in keys.unique():
-                group_values = values[keys == key].to_numpy()
-                statistics.append(stat_func(group_values))
-            return pd.Series(statistics)
+            df = pd.DataFrame({'keys': keys, 'values': values})
+            return df.groupby('keys')['values'].agg(stat_func)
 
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b2evqnt SDMetrics - Add metric for general MSAS statistics #639](https://app.clickup.com/t/86b2evqnt)

sdmetrics/timeseries/statistic_msas.py
```diff
@@ -0,0 +1,91 @@
...
+            'min': np.min,
+            'max': np.max,
+        }
+        if statistic not in statistic_functions:
+            raise ValueError(
+                f'Invalid statistic: {statistic}.'
+                f' Choose from [{", ".join(statistic_functions.keys())}].'
+            )
+
+        real_keys, real_values = real_data
```
amontanez24, line 77: should we check to make sure the inputs are tuples as well before trying to unpack?

sdmetrics/timeseries/statistic_msas.py
```diff
@@ -0,0 +1,91 @@
...
+        synthetic_keys, synthetic_values = synthetic_data
+        stat_func = statistic_functions[statistic]
+
+        def calculate_statistics(keys, values):
+            statistics = []
+            for key in keys.unique():
+                group_values = values[keys == key].to_numpy()
+                statistics.append(stat_func(group_values))
+            return pd.Series(statistics)
```
amontanez24, line 86: Is this faster than combining the two series into a dataframe and then using groupby with an [aggregation](https://pandas.pydata.org/docs/user_guide/groupby.html#built-in-aggregation-methods) function?

amontanez24 (approved):
:shipit:

frances-h (approved):
LGTM!

R-Palazzo (approved):

PR #654, opened 2024-11-07
Fix `KSComplement` instability for constant float values

### LINKED ISSUE #652: `KSComplement` can be unstable for constant float values
### Environment Details
* SDMetrics version: 0.16.0

### Error Description
The `KSComplement` may return a score of 0.0 when passed seemingly identical float values.

### Steps to reproduce

```python
from sdmetrics.single_column.statistical.kscomplement import KSComplement
import pandas as pd

vals1 = pd.Series([.2 - .1])
vals2 = pd.Series([.3 - .2])
print(KSComplement.compute(vals1, vals2)) # This prints 0.0 
```

### Expectation
The `KSComplement` is expected to return 1.0 for such cases.

### Additional context
This seems to be the default behavior of scipy's `ks_2samp`, which `KSComplement` uses to compute the KS test statistic.

fealho:
CU-86b2t0z0c, Resolve #652.

### DIFF
--- a/sdmetrics/single_column/statistical/kscomplement.py
+++ b/sdmetrics/single_column/statistical/kscomplement.py
@@ -1,5 +1,7 @@
 """Kolmogorov-Smirnov test based Metric."""
 
+import sys
+
 import numpy as np
 import pandas as pd
 from scipy.stats import ks_2samp
@@ -56,6 +58,13 @@ class KSComplement(SingleColumnMetric):
         if is_datetime(real_data):
             real_data = pd.to_numeric(real_data)
             synthetic_data = pd.to_numeric(synthetic_data)
+
+        try:
+            max_decimals = sys.float_info.dig - 1
+            real_data = real_data.round(max_decimals)
+            synthetic_data = synthetic_data.round(max_decimals)
+        except TypeError:
+            pass
 
         try:
             statistic, _ = ks_2samp(real_data, synthetic_data)
--- a/sdmetrics/single_column/statistical/kscomplement.py
+++ b/sdmetrics/single_column/statistical/kscomplement.py
@@ -59,12 +59,9 @@ class KSComplement(SingleColumnMetric):
             real_data = pd.to_numeric(real_data)
             synthetic_data = pd.to_numeric(synthetic_data)
 
-        try:
-            max_decimals = sys.float_info.dig - 1
-            real_data = real_data.round(max_decimals)
-            synthetic_data = synthetic_data.round(max_decimals)
-        except TypeError:
-            pass
+        max_decimals = sys.float_info.dig - 1
+        real_data = real_data.round(max_decimals)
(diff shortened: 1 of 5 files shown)


sdv-team:
Task linked: [CU-86b2t0z0c SDMetrics - `KSComplement` can be unstable for constant float values #652](https://app.clickup.com/t/86b2t0z0c)

sdmetrics/single_column/statistical/kscomplement.py
```diff
@@ -57,6 +59,13 @@ def compute(real_data, synthetic_data):
             real_data = pd.to_numeric(real_data)
             synthetic_data = pd.to_numeric(synthetic_data)
 
+        try:
```
fealho, line 62: The `try/catch` ensures invalid data raises the same error as before, ie it crashes during `ks_2samp`, not here. Without it various tests would need to be updated to use the error raised by `round` instead, which happens to be different for older versions of numpy and recent ones.

amontanez24: what are the situations where we can't round?

amontanez24: should we silently pass in those scenarios?

fealho: We could add more input validation, but that seems out of scope of this task, since it affects the behavior of the function besides what's described in the issue

fealho: Yes, it's just a little tricky because different errors show up for different numpy versions, so the tests need to take that into consideration. Let me know if that's what you prefer and I'll implement it.

amontanez24 (approved):

tests/integration/reports/multi_table/test_quality_report.py
```diff
@@ -299,6 +300,15 @@ def test_quality_report_with_errors():
         'Property': ['Column Shapes', 'Column Pair Trends', 'Cardinality', 'Intertable Trends'],
         'Score': [0.8165079365079364, 0.55, 0.95, 0.5833333333333334],
     })
+
+    pandas_version = version.parse(pd.__version__)
+    if pandas_version >= version.parse('2.2.0'):
```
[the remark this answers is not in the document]
fealho: Because different error messages are generated depending on the pandas version. For pandas < 2.2, the KSComplement crashes during rounding, for pandas >= 2.2 `round` doesn't crash anymore, instead the error is raised during `k2_samp`.

tests/integration/single_column/statistical/test_kscomplement.py
```diff
@@ -49,7 +49,12 @@ def test_bad(array_like):
 
 
 def test_one_float_value():
+    # Setup
```
R-Palazzo, line 52: You could add a docstring that describes the test here.

PR #662, opened 2024-11-14
Add `InterRowMSAS`, `StatisticMSAS` and `SequenceLengthSimilarity` metrics

### LINKED ISSUE #638: Add metric for sequence length similarity

### LINKED ISSUE #639: Add metric for general MSAS statistics

### LINKED ISSUE #640: Add metric for inter-row MSAS

### LINKED ISSUE #660: Fix `SequenceLengthSimilarity` docstrings
### Error Description
The docstrings for `SequenceLengthSimilarity`  include the following:

https://github.com/sdv-dev/SDMetrics/blob/6bec051b5532b69b6eef32631e76f15c7a95f08f/sdmetrics/timeseries/sequence_length_similarity.py#L43-L47

The type hints for the real and synthetic data parameters are incorrect. They should be `pd.Series`, not `Union[numpy.ndarray, pandas.DataFrame]`.

### LINKED ISSUE #661: Relocate timeseries metrics modules
### Error Description
`SequenceLengthSimilarity`, `InterRowMSAS` and `StatisticMSAS` are incorrectly located in the `sdmetrics.timeseries` folder:
  - `SequenceLengthSimilarity` should be moved to `sdmetrics.single_column.statistical`
  - `InterRowMSAS` and `StatisticMSAS` should be moved to `sdmetrics.column_pairs.statistical`

fealho:
resolves #640 
resolves #638
resolves #639
resolves #661 
resolves #660


amontanez24 (approved):
Can we do a rebase and merge @fealho

frances-h (approved):

PR #678, opened 2024-12-02
Add `DisclosureProtection` metric

### LINKED ISSUE #675: Add `DisclosureProtection` metric
### Problem Description
The [CategoricalCAP metric](https://docs.sdv.dev/sdmetrics/metrics/metrics-glossary/categoricalcap) is currently available in SDMetrics. It is based on a well-researched and cited methodology for measuring the risk of data disclosure. It acts as a great measure of privacy for particular sensitive columns.

However, there are a number of issues which make this metric hard for users to use and interpret. We'd like to create a new metric that wraps around the CategoricalCAP metric and provides some much needed quality-of-life improvements.

### Expected behavior
The `DisclosureProtection` metric should wrap around the CategoricalCAP metric with some ease-of-use features:
- It considers nulls to be a single category
- It has the ability to discretize continuous columns
- It exposes all different variants of CAP under a single parameter
- It returns a score that is normalized to a baseline of random data (available as a breakdown)

#### Parameters:
- `real_data (pd.DataFrame)`: A pandas DataFrame with the real data.
- `synthetic_data (pd.DataFrame)`: A pandas DataFrame with the synthetic data.
- `known_column_names (list[str])`: A list of names of the columns that the attacker may already know.
- `sensitive_column_names (list[str])`: A list of the names of the columns that an attacker wants to guess.
- `computation (str, optional)`: The type of computation we'll use to simulate the attack. Must be one of `['cap', 'generalized_cap', 'zero_cap']`. Defaults to `'cap'`. 
- `continuous_column_names (list[str], optional)`: A list of column names that represent continuous values. These columns will be discretized. Defaults to `None`.
- `num_discrete_bins (int, optional)`: The number of bins to create when discretizing continuous columns. Defaults to 10 bins.

#### Computation:
Internally, metric should first compute a baseline score that assumes random data. Note that we do not actually have to create random data to compute this; it is enough to know the # of unique categories in the sensitive column. (If there are multiple sensitive columns, then the unique_category_count is the product of the counts of each individual column.)
`baseline_protection = 1 - 1/(unique_category_count)`

Then the metric will compute the CAP protection score by running the appropriate metric requested (CategoricalCAP, CategoricalZeroCAP, or CategoricalGeneralizedCAP). Call this score cap_protection
`cap_protection = CategoricalCAP.compute(...)`

The final returned score should be:
`score = min(CAP_protection/baseline_protection, 1 ).`

#### Compute Breakdown:
Additionally, we should provide `compute_breakdown` functionality which will return the breakdown of how the DisclosureProtection metric was calculated. It should return a dictionary containing the `'score'` for the metric, the `'cap_protection'` which is the score from the underlying CAP metric being used, and the `'baseline_protection'` which is the CAP score from using entirely random data. 

```python
from sdmetrics.single_table import DisclosureProtection

DisclosureProtection.compute_breakdown(
    real_data=real_table,
    synthetic_data=synthetic_table,
    known_columns=['age', 'gender'],
    sensitive_column=['political_affiliation'],
    columns_to_discretize=['age'],
)
```
```
{
    'score': 0.912731436159061,
    'cap_protection': 0.782341231,
    'baseline_protection': 0.85714285715
}
```

### Additional context
We should also add a warning to the CategoricalCAP, CategoricalGeneralizedCAP, and CategoricalZeroCAP metrics to recommend using the DataProtection metric instead. This warning should only appear when a user tries to run the metric directly.
```
>>> CategoricalCAP.compute(...)
Warning: Consider using the new 'DisclosureProtection' metric that provides extra functionality on top of CategoricalCAP
```

frances-h:
Resolve #675
CU-86b2ymek7

### DIFF
--- a/sdmetrics/single_table/privacy/disclosure_protection.py
+++ b/sdmetrics/single_table/privacy/disclosure_protection.py
@@ -85,9 +85,13 @@ class DisclosureProtection(SingleTableMetric):
     @classmethod
     def _discretize_column(cls, real_column, synthetic_column, num_bins):
         bin_labels = [str(x) for x in range(num_bins)]
-        real_binned, bins = pd.cut(real_column, num_bins, labels=bin_labels, retbins=True)
+        real_binned, bins = pd.cut(
+            pd.to_numeric(real_column.to_numpy()), num_bins, labels=bin_labels, retbins=True
+        )
         bins[0], bins[-1] = -np.inf, np.inf
-        synthetic_binned = pd.cut(synthetic_column, bins, labels=bin_labels)
+        synthetic_binned = pd.cut(
+            pd.to_numeric(synthetic_column.to_numpy()), bins, labels=bin_labels
+        )
 
         return real_binned.to_numpy(), synthetic_binned.to_numpy()
 
--- a/tests/unit/single_table/privacy/test_disclosure_protection.py
+++ b/tests/unit/single_table/privacy/test_disclosure_protection.py
@@ -115,11 +115,12 @@ class TestDisclosureProtection:
         # Assert
         assert null_category_map == {'col1': '__NULL_VALUE___', 'col2': '__NULL_VALUE___'}
 
-    def test__discreteize_column(self):
+    @pytest.mark.parametrize('dtype', ['int32', 'int64', 'Int32', 'Int64'])
+    def test__discretize_column_int_dtypes(self, dtype):
         """Test discretizing a continous column"""
         # Setup
-        real_column = pd.Series([0, 2, 6, 8, 10])
-        synthetic_column = pd.Series([-10, 1, 3, 5, 7, 9, 20])
+        real_column = pd.Series([0, 2, 6, 8, 10], dtype=dtype)
+        synthetic_column = pd.Series([-10, 1, 3, 5, 7, 9, 20], dtype=dtype)
 
         # Run
         binned_real, binned_synthetic = DisclosureProtection._discretize_column(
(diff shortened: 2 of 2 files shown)


sdv-team:
Task linked: [CU-86b2ymek7 SDMetrics - Add `DisclosureProtection` metric #675](https://app.clickup.com/t/86b2ymek7)

tests/integration/single_table/privacy/test_disclosure_protection.py
```diff
@@ -0,0 +1,144 @@
...
+            good_synthetic_data,
+            sensitive_column_names=sensitive_columns,
+            known_column_names=known_columns,
+            continuous_column_names=continuous_columns,
+            num_discrete_bins=10,
+        )
+
+        # Assert
+        assert score_breakdown == {
+            'score': 0.8979591836734694,
```
gsheni, line 91: Should we round this?

tests/integration/single_table/privacy/test_disclosure_protection.py
```diff
@@ -0,0 +1,144 @@
...
+            bad_synthetic_data,
+            sensitive_column_names=sensitive_columns,
+            known_column_names=known_columns,
+            continuous_column_names=continuous_columns,
+            num_discrete_bins=10,
+        )
+
+        # Assert
+        assert score_breakdown == {
+            'score': 0.40816326530612246,
```
gsheni, line 115: Same question, should we round?

sdmetrics/single_table/privacy/disclosure_protection.py
```diff
@@ -0,0 +1,241 @@
...
+                null_value += '_'
+
+            null_category_map[col] = null_value
+
+        return null_category_map
+
+    @classmethod
+    def _discretize_column(cls, real_column, synthetic_column, num_bins):
+        bin_labels = [str(x) for x in range(num_bins)]
+        real_binned, bins = pd.cut(real_column, num_bins, labels=bin_labels, retbins=True)
```
gsheni, line 88: What happens if the real_column contains NaNs? It seems to work
```python
import pandas as pd
import numpy as np
real_binned, bins = pd.cut(pd.Series([1, 7, 5, 4, 6, pd.NA], dtype='Int64'), 3, retbins=True)
real_binned
bins
```

```python
In [17]: real_binned
Out[17]:
0    (0.994, 3.0]
1      (5.0, 7.0]
2      (3.0, 5.0]
3      (3.0, 5.0]
4      (5.0, 7.0]
5             NaN
dtype: category
Categories (3, interval[float64, right]): [(0.994, 3.0] < (3.0, 5.0] < (5.0, 7.0]]

In [18]: bins
Out[18]: array([0.994, 3.   , 5.   , 7.   ])
```

tests/unit/single_table/privacy/test_disclosure_protection.py
```diff
@@ -0,0 +1,254 @@
...
+        CAPMethodsMock.keys.return_value = ['CAP', 'ZERO_CAP', 'GENERALIZED_CAP']
+        CAPMethodsMock.get.return_value = CAPMock
+
+        # Run
+        score_breakdown = DisclosureProtection.compute_breakdown(
+            real_data=real_data,
+            synthetic_data=synthetic_data,
+            known_column_names=['col1'],
+            sensitive_column_names=['col2', 'col3'],
+            continuous_column_names=['col3'],
```
gsheni, line 176: Should we verify that we support all dtypes with `continuous_column_names`? Float64, Float32, Int64, Int32 are some additional ones to look into (beyond float64 and int64 (numpy dtypes)).

sdmetrics/single_table/privacy/disclosure_protection.py
frances-h: Yeah this is expected, we handle all null values separately after we discretize all the columns.

gsheni: Gotcha, sounds good

pvk-developer (approved):
LGTM!

tests/integration/single_table/privacy/test_disclosure_protection.py
frances-h: Do you mean for the test or the return value itself? We don't usually round the scores returned from metrics.

gsheni: Gotcha. I was referencing the metric score

gsheni (approved):

PR #714, opened 2025-02-04
Add `BinaryClassifierPrecisionEfficacy` metric

### LINKED ISSUE #711: Add `BinaryClassifierPrecisionEfficacy` metric
### Problem Description

There are many ways and reasons to perform data augmentation with synthetic data for the purposes for building ML models. While we have some ML Efficacy metrics in beta, we'd like to create a suite of metrics that more effectively cover the use case. The `BinaryClassifierPrecisionEfficacy` metric will specifically measure if synthetic data improves the precision of a binary classifier.

### Expected behavior
This metric should be defined in the `data_augmentation` sub-module inside `single_table`.
```python
from sdmetrics.single_table.data_augmentation import BinaryClassifierPrecisionEfficacy

BinaryClassifierPrecisionEfficacy.compute_breakdown(
  real_training_data=real_df,
  synthetic_data=synthetic_df,
  real_validation_data=real_holdout_df,
  prediction_column_name='covid_status',
  minority_class_label=1,
  classifier='XGBoost',
  fixed_recall_value=0.9
)
```

#### `compute_breakdown`
**API**
- Args
  - `real_training_data (pd.DataFrame)` - A dataframe containing the real data that was used for training the synthesizer. The metric will use this data for training a Binary Classification model.
  - `synthetic_data (pd.DataFrame)` - A dataframe containing the synthetic data sampled from the synthesizer. The metric will use this data for training a Binary Classification model. 
  - `real_validation_data (pd.DataFrame)` - A dataframe containing a holdout set of real data. This data should not have been used to train the synthesizer. This data will be used to evaluate a Binary Classification model. 
  - `metadata (dict)` - The metadata dictionary describing the table of data.
  - `prediction_column_name (str)` - The name of the column to be predicted. The column should be a categorical or boolean column.
  - `minority_class_label [str/int/float]` - The value in the prediction column that should be considered a positive result, from the perspective of Binary Classfication. All other values in the column will be considered negative results. 
  - `classifier [str, optional]` - The ML algorithm to use when building a Binary Classfication. Supported options are 'XGBoost'. Defaults to 'XGBoost'.
    - Note: as an MVP, we will only support XGBoost. Future feature requests may add support for additional algorithms. 
  - `fixed_recall_value [float, optional]` - A float in the range (0, 1.0) describing the value to fix for the recall when building the Binary Classification model. Defaults to 0.9.
- Returns
  - A dictionary of the breakdown of the score, with the following information:
    - The score for the metric. This is the improvement precision score (from baseline -> augmented data) in percentage points, `score = MIN(0, augmented_precision_score - baseline_precision_score))`. 
    - The parameters used to run the metric
    - For each of the augmented data and the real data baseline:
      - The recall score achieved during training. This should be at least the requested score input as a parameter, but may not be exactly equal.
      - The actual recall score achieved on the validation (holdout) set.
      - The precision score achieved on the validation set.
      - The prediction counts achieved on the validation set (true positive, false positive, true negative, and false negative).
  - Expected dictionary output:  
  ```python
  {
    'score': 0.86,
    'augmented_data': {
      'recall_score_training': 0.950,
      'recall_score_validation': 0.912
      'precision_score_validation': 0.84,
        'true_positive': 21,
        'false_positive': 4,
        'true_negative': 73,
        'false_negative': 3
      },
    },
    'real_data_baseline': {
      # keys are the same as the 'augmented_data' dictionary
     },
    'parameters': {
      'prediction_column_name': 'covid_status',
      'minority_class_label': 1,
      'classifier': 'XGBoost',
      'fixed_recall_value': 0.9
    }
  }
  ```
**Algorithm**
1. Concatanate the `real_training_data` and `synthetic_data together`
2. Train a binary classification model on the data, using the classifier algorithm selected (default: XGBoost)
  a) Need to pre-process the data to turn discrete columns into continuous columns (note that we cannot use RDT, and should use scikit learn methods instead)
  b) Data pre-processing to convert the `prediction_columm` into a boolean column with the correct positive/negative values
    - If multi-class, consider only the `minority_class_label` as positive values. All other values will be considered negative.
3. Based on the parameters, fix the recall for the minority class
  a)  This will require finding the right threshold to achieve as close of the fixed recall as specified. The classifier will return a continuous value for each data point in training data and we would have to find the threshold that will achieve the value closest to the fixed rate. **Note that we should always choose a threshold that is as close as possible to the requested recall value but never less than it. That is to say, ensure that the training set recall is >= the requested recall value.**
  b) Save this threshold to use on the validation data. This threshold is now a learnt parameter alongside the classifier
4. Take the classifier and apply it on the `real_validation_data`. Compute the statistics that we want to return.
5. Calculate the baseline. Repeat steps 1-4 but this time, only use the `real_training_data` (do not concatenate `synthetic_data`).

#### `compute`

The `compute` method should return just the overall `score` parameter calculated by `compute_breakdown`.

### Additional context
See [this doc](https://docs.google.com/document/d/1HS1bJRwjbp8XA58E54YOR9zTUBUyGBmABXgDq8VSl9g/edit?usp=sharing)

There will be significant overlap of required pre-processing/helper functions between data augmentation metrics. When possible, general functionality should be abstracted into utility functions that can be reused across many metrics.

R-Palazzo:
CU-86b3m9u4d
Resolve #711


sdv-team:
Task linked: [CU-86b3m9u4d SDMetrics - Add `BinaryClassifierPrecisionEfficacy` metric #711](https://app.clickup.com/t/86b3m9u4d)

sdmetrics/single_table/data_augmentation/utils.py
```diff
@@ -0,0 +1,126 @@
...
+            ' to a dictionary.'
+        )
+
+    if not isinstance(prediction_column_name, str):
+        raise TypeError('`prediction_column_name` must be a string.')
+
+    if classifier is not None and not isinstance(classifier, str):
+        raise TypeError('`classifier` must be a string or None.')
+
+    if classifier != 'XGBoost':
```
amontanez24, line 36: For these types of validations I think it's good to split them into separate smaller functions. The reason for this is that if we decide to expand this class to support more classifiers or create a new class to do that, we might want to overwrite just this part of the validation and keep everything else.

sdmetrics/single_table/data_augmentation/utils.py
```diff
@@ -0,0 +1,126 @@
...
+            f'is not present in the column `{prediction_column_name}` for the real validation data.'
+            ' The `precision`and `recall` are undefined for this case.'
+        )
+
+    synthetic_labels = set(synthetic_data[prediction_column_name].unique())
+    real_labels = set(real_training_data[prediction_column_name].unique())
+    if not synthetic_labels.issubset(real_labels):
+        raise ValueError(
+            f'The ``{prediction_column_name}`` column must have the same values in the real '
+            'and synthetic data. The synthetic data has the following unseen values: '
```
amontanez24, line 94:
```suggestion
            'and synthetic data. The following values are present in the synthetic data and not the real data: '
```

sdmetrics/single_table/data_augmentation/base.py
```diff
@@ -0,0 +1,258 @@
...
+            f'{self._metric_to_fix}_score_training': training_score,
+            'recall_score_validation': recall,
+            'precision_score_validation': precision,
+            'prediction_counts_validation': prediction_counts_validation,
+        }
+
+        return scores
+
+    @classmethod
+    def compute_breakdown(
```
amontanez24, line 145: why is this a class method?

sdmetrics/single_table/data_augmentation/base.py
```diff
@@ -0,0 +1,258 @@
...
+
+        All the preprocessing steps are applied to the tables before applying the
+        data augmentation technique.
+
+        Args:
+            tables (dict[str, pandas.DataFrame]):
+                The tables to transform.
+        """
+        tables_result = {}
+        for table_name, table in tables.items():
```
amontanez24, line 68: Isn't this a single table metric?

sdmetrics/single_table/data_augmentation/base.py
```diff
@@ -0,0 +1,258 @@
...
+        self.fixed_value = fixed_value
+        # To assess the preicision efficacy, we have to fix the recall and reciprocally
+        self._metric_to_fix = 'recall' if self.metric_name == 'precision' else 'precision'
+        self._metric_method = METRIC_NAME_TO_METHOD[self._metric_to_fix]
+        self._classifier_name = classifier
+        self._classifier = XGBClassifier(enable_categorical=True)
+
+        self._fit_preprocess(data, metadata)
+
+    def _transform_preprocess(self, tables):
```
amontanez24, line 57: minor: we could just call transform

sdmetrics/single_table/data_augmentation/base.py
```diff
@@ -0,0 +1,258 @@
...
+            tables_result[table_name] = table
+
+        return tables_result
+
+    def _get_best_threshold(self, train_data, train_target):
+        """Find the best threshold for the classifier model."""
+        target_probabilities = self._classifier.predict_proba(train_data)[:, 1]
+        precision, recall, thresholds = precision_recall_curve(train_target, target_probabilities)
+        # To assess the preicision efficacy, we have to fix the recall and reciprocally
+        metric = precision if self.metric_name == 'recall' else recall
```
amontanez24, line 86: can't you just make this self._metric_to_fix?

sdmetrics/single_table/data_augmentation/base.py
```diff
@@ -0,0 +1,258 @@
...
+            real_validation_data,
+            metadata,
+            prediction_column_name,
+            minority_class_label,
+            classifier,
+            fixed_recall_value,
+        )
+        result = {}
+        metric = cls()
+        metric._fit(
```
amontanez24, line 169: something about this seems strang4e. Maybe we should make a separate class to represent the object that fits/trains the classifier and makes predictions. Then this metric can just have a reference to that object

sdmetrics/single_table/data_augmentation/base.py
R-Palazzo: I find `_fit_preprocess` and `_transform_preprocess` clearer, and there is no confusion with the `fit` method of the ML model. The ML model does not have a transform method, but since we keep `_fit_preprocess`, it's consistent to name `_transform_preprocess` as well.

sdmetrics/single_table/data_augmentation/base.py
R-Palazzo: It's a class method by definition:
https://github.com/sdv-dev/SDMetrics/blob/280a93e034a33d050f47ddf576b4efb63505f1e8/sdmetrics/single_table/base.py#L168

R-Palazzo:
Thanks for the review @amontanez24. I addressed the comments in [791c7d5](https://github.com/sdv-dev/SDMetrics/pull/714/commits/791c7d5599545c923ac24c94518fdf3bf0ec09da) and [aafdd1e](https://github.com/sdv-dev/SDMetrics/pull/714/commits/aafdd1e1abc48732abb92c475585fe02af1dcd79) with some refactoring as you recommended

sdmetrics/visualization.py
```diff
@@ -32,7 +32,7 @@ def wrapper(*args, **kwargs):
             ipython_interpreter = str(get_ipython())
             if 'ZMQInteractiveShell' in ipython_interpreter and 'iframe' in renderers:
                 # This means we are using jupyter notebook
-                pio.renderers.default = 'vscode'
+                pio.renderers.default = 'iframe'
```
R-Palazzo, line 35: I found that this change was accidentally added in a previous PR, so I reverted it.

sdmetrics/single_table/data_augmentation/base.py
amontanez24: The ML model was separated out though. I think _fit_preprocess confuses me as it makes me wonder if we are fitting on preprocessed data or fitting as part of the preprocess step

amontanez24: @R-Palazzo

R-Palazzo: @amontanez24 sorry I missed this, done in [fbc5d37](https://github.com/sdv-dev/SDMetrics/pull/714/commits/fbc5d372879c7313dc46a13543158e6a629a8110)

amontanez24 (approved):
LGTM! @frances-h Can you take a look? You know the details of the algorithm better than I do

frances-h (approved):

PR #722, opened 2025-02-11
Allow subsampling when computing the `ContingencySimilarity` metric

### LINKED ISSUE #716: Allow subsampling when computing the `ContingencySimilarity` metric
### Problem Description
The [ContingencySimilarity metric](https://docs.sdv.dev/sdmetrics/metrics/metrics-glossary/contingencysimilarity) computes an entire contingency table for both the real and synthetic data in order to compare the discrete, 2D distributions. Our experiments have shown that for large datasets with high cardinality (i.e. large # of category values), this metric is not very performant.

Our experiments have also shown that a simple approach of subsampling the data (both the real dataset and the synthetic dataset) can yield much faster performance without affecting the final score by too much (within 5%). Based on this, we should add a parameter to this metric to allow for subsampling.

### Expected behavior
Add an optional parameter to `ContingencySimilarity` called `num_rows_subsample`
- (default) `None`: Do not subsample the rows
- `<integer>`: Randomly subsample the provided number of rows for both the real and the synthetic datasets before computing the metric

```python
from sdmetrics.column_pairs import ContingencySimilarity

ContingencySimilarity.compute(
    num_rows_subsample=1000
)
```

### Additional context
Our experiments have shown that multiple iterations are not needed when doing such a subsample, as the overall score is not affected by much. So we are not adding any parameter for iterations.

R-Palazzo:
CU-86b3v4dm3
Resolve #716

### DIFF
--- a/sdmetrics/column_pairs/statistical/contingency_similarity.py
+++ b/sdmetrics/column_pairs/statistical/contingency_similarity.py
@@ -27,7 +27,9 @@ class ContingencySimilarity(ColumnPairsMetric):
     max_value = 1.0
 
     @staticmethod
-    def _validate_inputs(real_data, synthetic_data, continuous_column_names, num_discrete_bins):
+    def _validate_inputs(
+        real_data, synthetic_data, continuous_column_names, num_discrete_bins, num_rows_subsample
+    ):
         for data in [real_data, synthetic_data]:
             if not isinstance(data, pd.DataFrame) or len(data.columns) != 2:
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b3v4dm3 SDMetrics - Allow subsampling when computing the `ContingencySimilarity` metric #716](https://app.clickup.com/t/86b3v4dm3)

fealho (approved):
Looks good 👍

frances-h (approved):

PR #725, opened 2025-02-18
Update final scoring method for `BinaryClassifierEfficacy` precision & recall metrics (make 0.5 a threshold)

### LINKED ISSUE #724: Update final scoring method for `BinaryClassifierEfficacy` precision & recall metrics (make 0.5 a threshold)
### Problem Description
In #712 and #713 we implemented new metrics to help users understand whether augmenting real data with synthetic data is will help with binary classification. The computation computed the binary classification score for just the real data (`baseline_score`) and the augmented data with real + synthetic data (`augmented_score`). The final score computation was:

```
score = MIN(0, augmented_score - baseline_score))
```

This scoring system meets the SDMetrics requirements that:
- The final score must be in the range [0, 1]
- 0 must correspond to the worse outcome and 1 must correspond to the best outcome

With this scoring system, any score >0 means that synthetic data is _helping_ the binary classification problem -- which is a good thing. So, for eg. even a score of 0.1 is considered a good thing because it means synthetic data is yielding a +10% improvement in classification. This is different from other metrics, where scores closer to 0 are considered to be bad. For example in all of our quality metrics, a score of 0.1 indicates bad quality synthetic data.

### Expected behavior
Update the scoring method for both `BinaryClassifierPrecisionEfficacy` and `BinaryClassifierRecallEfficacy` metrics.

```
score = (augmented_score - baseline_score)/2 + 0.5
```

This scoring system has a different interpretation that is more inline with other metrics:

- `score < 0.5` indicates that synthetic data is _harming the binary classification_, which is a bad thing
- `score = 0.5` indicates that the synthetic data is not having any effect on the binary classification, which is netural
- `score > 0.5` indicates that the synthetic data is _improving the binary classification_, which is a good thing.
 
Note that the augmented and baseline precision/recall scores are each in the range [0,1]. So based on this formula, the overall score will continue to be range [0, 1].

frances-h:
Resolve #724 
CU-86b3y6b2n


sdv-team:
Task linked: [CU-86b3y6b2n SDMetrics - Update final scoring method for `BinaryClassifierEfficacy` precision & recall metrics (make 0.5 a threshold) #724](https://app.clickup.com/t/86b3y6b2n)

amontanez24 (approved):

R-Palazzo (approved):
LGTM!
Could you add the defaults values of the `binary classifier preicision efficacy` metric in this PR?
https://github.com/sdv-dev/SDMetrics/blob/7261608c5d7d24e3e2acd1c3126dabbdaa339867/sdmetrics/single_table/data_augmentation/binary_classifier_precision_efficacy.py#L21

PR #721, opened 2025-02-10
Support python 3.13

rwedge:
Resolves #645 

CU-86b2mqatx

Issue #720, opened 2025-02-10
Add `DCRBaselineProtection` metric


frances-h:
### Problem Description
Currently in SDMetrics, we are offering [DisclosureProtection](https://docs.sdv.dev/sdmetrics/metrics/metrics-glossary/disclosureprotection) as the only metric for measuring data privacy. We've vetted this metric to make sure that it's understandable and insightful. However, running this metric requires the user to make some decisions about what to input as known columns and sensitive columns. This means that the privacy score is only ever based on a subset of columns.

We'd like to create a new metric that could be run on an entire set of columns.

### Expected behavior
Add the `DCRBaselineProtection` metric. This metric uses a DCR (distance to closest record) computation to measure how close the synthetic data is to the real data – as opposed to a baseline of random data. 

```python
from sdmetrics.single_table import DCRBaselineProtection

DCRBaselineProtection.compute(
  real_training_data=real_df,
  synthetic_data=synthetic_df,
  metadata=my_metadata)

DCRBaselineProtection.compute_breakdown(
  real_training_data=real_df,
  synthetic_data=synthetic_df,
  metadata=my_metadata)
```

#### `compute_breakdown`
__Args__
- `real_training_data (pd.DataFrame)`: A pandas.DataFrame object containing the real data used for training the synthesizer.
- `synthetic_data (pd.DataFrame)`: A pandas.DataFrame object containing the synthetic data sampled from the synthesizer.
- `metadata (dict)`: A metadata dictionary that describes the table of data.
- `num_rows_subsample (int, optional)`: The number of synthetic data rows to subsample from the synthetic data. Use this to increase the speed of the computation, if your dataset is large. Default is `None` (do not subsample the data).
- `num_iterations (int, optional)`: The number of iterations to perform when subsampling. The final score will be the average of all iterations. Default is 1.

__Returns__
A dictionary breakdown of the score, with the following information:
- The overall score of the metric
- The median DCR score between the synthetic data and real data, and the median DCR score between the random data and real data.
  - In case of multiple iterations, the median DCR in the breakdown should be the averages across all iterations.

Example
```python
{
  'score': 0.58808,
  'median_DCR_to_real_data': {
    'synthetic_data': 0.12355,
    'random_data_baseline': 0.21009
  }
}
```

__Computation Algorithm__
[...]

PR #726, opened 2025-02-19
Add DCR measurement functions

lajohn4747:
This is the core DCR algorithm as noted in #720 and #721

Added unit tests to confirm the DCR algorithm works as expected.


sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+            col_range = datetime_to_timestamp_col.max() - datetime_to_timestamp_col.min()
+        column_ranges[column] = col_range
+
+    dcr_dist_dict = synthetic_data.apply(
+        lambda synth_row: _calculate_dcr_between_row_and_data(
+            synth_row, comparison_data, column_ranges, metadata
+        ),
+        axis=1,
+    )
```
lajohn4747, line 160: I used `apply` as it considered optimal for performance rather than doing other iteration techniques such as `iterrows` and `itertuples`

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,162 @@
+"""Distance to closest record measurement functions."""
+
+import pandas as pd
+
+from sdmetrics.utils import is_datetime
+
+
+def _calculate_dcr_value(s_value, d_value, col_name, metadata, range=None):
```
frances-h, line 8: Can we rename `range` to something like `col_range` instead since `range` conflicts with the python function?

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    Arguments:
+        s_value (int, float, datetime, boolean, string, or None):
+            The synthetic value that we are calculating DCR value for
+        d_value (int, float, datetime, boolean, string, or None):
+            The data value that we are referencing for measuring DCR.
+        col_name (string):
+            The column name in the metadata dictionary that will be used to
+            determine how the type is measured.
+        metadata (dict):
+            The metadata dict.
```
frances-h, line 20: Could we just pass in the column sdtype instead of the column name and entire metadata?

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Returns:
+        pandas.Dataframe:
+            Returns a dataframe that shows the DCR value for all synthetic data.
+    """
+    column_ranges = {}
+    for column in comparison_data.columns:
+        if column not in metadata['columns']:
+            raise ValueError(f'Column {column} was not found in the metadata.')
```
frances-h, line 133: Add a newline below

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+            if not is_datetime(datetime_to_timestamp_col):
+                datetime_to_timestamp_col = pd.to_datetime(
+                    datetime_to_timestamp_col, format=datetime_format, errors='coerce'
+                )
+
+            if not isinstance(datetime_to_timestamp_col, pd.Timestamp):
+                datetime_to_timestamp_col = datetime_to_timestamp_col.apply(
+                    lambda x: x.timestamp() if pd.notna(x) else x
+                )
```
frances-h, line 150: Do we need this because we're coercing errors above? It might be better to let `pd.to_datetime` error out in that case, we can add a warning if the metadata doesn't have a datetime format specified for the datetime columns.

lajohn4747: Sounds good, added a warning.

tests/unit/single_table/privacy/test_dcr_utils.py
```diff
@@ -0,0 +1,449 @@
...
+    calculate_dcr,
+)
+from tests.utils import check_if_value_in_threshold
+
+
+@pytest.fixture()
+def real_data():
+    return pd.DataFrame({
+        'num_col': [10, 20, np.nan, 40, 50, 60],
+        'cat_col': ['A', 'B', 'A', None, 'B', 'C'],
```
gsheni, line 23: What about categorical columns as integers?

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        if pd.isna(s_value) and pd.isna(d_value):
+            return 0.0
+        elif pd.isna(s_value) or pd.isna(d_value):
+            return 1.0
+        else:
+            distance = abs(s_value - d_value) / (col_range)
+            return min(distance, 1.0)
+    else:
+        if s_value == d_value:
+            return 0.0
```
R-Palazzo, line 42: What about the comparison of `nans` here? We might want to convert them to None so that the values can be compared.

R-Palazzo (commented):
Good job!
Should we create a feature branch that all related PRs to the DCR measurement can point to?

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,168 @@
...
+            The synthetic value that we are calculating DCR value for
+        d_value (int, float, datetime, boolean, string, or None):
+            The data value that we are referencing for measuring DCR.
+        sdtype (string):
+            The sdtype of the column values.
+        col_range (float):
+            The range of values for a column used for numerical values to calculate DCR.
+
+    Returns:
+        pandas.Dataframe:
```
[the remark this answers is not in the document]
lajohn4747: Yep you're right, updated.

sdmetrics/single_table/privacy/dcr_utils.py
lajohn4747: Numerical and datetime should be handled with `pd.na` in the above code. This branch is for categorical so if there is a None it would be considered a different category I believe.

lajohn4747: Update to move the none checking to be outside the `numerical`/`datetime` check so it should also handle the nans in categorical data.

tests/unit/single_table/privacy/test_dcr_utils.py
lajohn4747: Yep added another column to these tests!

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,172 @@
...
+            The metadata dict.
+
+    Returns:
+        pandas.Dataframe:
+            Returns a dataframe that shows the DCR value for all synthetic data.
+    """
+    column_ranges = {}
+    missing_cols = set(real_data.columns) - set(synthetic_data.columns)
+    if missing_cols:
+        raise ValueError(f'Different columns detected: {missing_cols}')
```
frances-h, line 145: We should check the columns against the metadata first, I think it's okay if there's extra columns in one or the other that we just ignore because they aren't listed in the metadata. We can also check the real/synthetic data have all expected columns from the metadata, and warn if they're missing any. Then during calculation, just only check the columns that are present in both (but should still error if there's no overlapping statistical columns).

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,172 @@
...
+
+    real_data_copy = real_data.copy()
+    synthetic_data_copy = synthetic_data.copy()
+    _convert_datetime(real_data_copy, metadata)
+    _convert_datetime(synthetic_data_copy, metadata)
+
+    for column in real_data_copy.columns:
+        if column not in metadata['columns']:
+            raise ValueError(f'Column {column} was not found in the metadata.')
```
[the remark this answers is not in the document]
lajohn4747: I rebased and used the new function `_process_data_with_metadata` to handle the metadata intersections.

sdmetrics/single_table/privacy/dcr_utils.py
lajohn4747: I rebased and used the new function `_process_data_with_metadata` to handle the metadata intersections. I also wrote a test to confirm only overlapping columns are measured..

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+
+        difference = abs(synthetic_value - real_value)
+        if isinstance(difference, pd.Timedelta):
+            difference = difference.total_seconds()
+
+        distance = difference / col_range
+        return min(distance, 1.0)
+
+    else:
```
frances-h, line 46: Can remove the else since we return in the `if`

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -0,0 +1,154 @@
...
+            The metadata dict.
+
+    Returns:
+        float:
+            Returns the minimum distance to closest record computed between the
+            synthetic row and the reference dataset.
+    """
+    dist_srow_to_all_rows = real_data.apply(
+        lambda d_row_obj: _calculate_dcr_between_rows(
+            synthetic_row, d_row_obj, column_ranges, metadata
```
frances-h, line 102: minor: can we rename `d_row_obj` to `real_row`?

frances-h (commented):
Looking good! Just some minor cleanup feedback

R-Palazzo (approved):

PR #728, opened 2025-02-19
Add DCRBaselineProtection Metric

### LINKED ISSUE #720: Add `DCRBaselineProtection` metric

lajohn4747:
resolves #720 
CU-86b3w869f

Average of 84 secs to run a synthetic sample of size 1000 to run against the demo dataset `fake_hotels_guest` (400 rows used for training and 100 rows used for validation) without subsampling.


sdv-team:
Task linked: [CU-86b3w869f SDMetrics - Add `DCRBaselineProtection` metric #720](https://app.clickup.com/t/86b3w869f)

sdmetrics/single_table/privacy/dcr_baseline_protection.py
```diff
@@ -0,0 +1,199 @@
...
+            sanitized_synthetic_data,
+            sanitized_real_validation_data,
+        )
+
+    @classmethod
+    def compute_breakdown(
+        cls,
+        real_training_data,
+        synthetic_data,
+        real_validation_data,
```
frances-h, line 84: The DCRBaselineProtection metric shouldn't take in any validation data. Per the issue, it should just take in real and synthetic data and then generate fake random data to use as a comparison.

lajohn4747: Talked offline about the confusion of using validation data. Updated the doc to remove references to validation data in DCRBaselineProtection.

sdmetrics/single_table/privacy/dcr_baseline_protection.py
```diff
@@ -0,0 +1,199 @@
...
+from sdmetrics.single_table.base import SingleTableMetric
+from sdmetrics.single_table.privacy.dcr_utils import calculate_dcr
+from sdmetrics.utils import get_columns_from_metadata, get_type_from_column_meta
+
+
+class DCRBaselineProtection(SingleTableMetric):
+    """DCR Baseline Protection metric."""
+
+    name = 'DCRBaselineProtection'
+    goal = Goal.IGNORE
```
R-Palazzo, line 13: Should it be `Goal.MAXIMIZE`? Since the best score is 1.0

tests/integration/single_table/privacy/test_dcr_baseline_protection.py
```diff
@@ -0,0 +1,265 @@
...
+        compute_num_iteration_1 = DCRBaselineProtection.compute_breakdown(
+            train_data, synthetic_data, holdout_data, metadata, num_rows_subsample, 1
+        )
+        compute_num_iteration_1000 = DCRBaselineProtection.compute_breakdown(
+            train_data, synthetic_data, holdout_data, metadata, num_rows_subsample, 1000
+        )
+        compute_train_same = DCRBaselineProtection.compute_breakdown(
+            synthetic_data, synthetic_data, holdout_data, metadata, num_rows_subsample, 1000
+        )
+
```
R-Palazzo, line 263: Add `# Assert`

R-Palazzo (commented):
Looks good!
Could you add some unit tests?

As mentioned in https://github.com/sdv-dev/SDMetrics/pull/726, it will be nice to use as much as possible metadata utils method when possible. For instance:
- Checking modelable sdtypes
- Converting datetime
- Filtering out columns

We start defning them in https://github.com/sdv-dev/SDMetrics/pull/729.
Feel free to add some methods specific to the DCR metrics. Then it's easier to maintain if we update something regarding the metadata.

sdmetrics/single_table/privacy/dcr_baseline_protection.py
```diff
@@ -0,0 +1,199 @@
...
+            raise ValueError('num_iterations must be an integer.')
+
+        if num_iterations < 1:
+            raise ValueError(f'num_iterations ({num_iterations}) must be greater than 1.')
+
+        if metadata is not None:
+            if not isinstance(metadata, dict):
+                metadata = metadata.to_dict()
+
+        valid_sdtypes = {'numerical', 'categorical', 'boolean', 'datetime'}
```
amontanez24, line 49: @R-Palazzo I think you define this again in one of your PRs. Can we move this list to a centralized location?

sdmetrics/single_table/privacy/util.py
```diff
@@ -148,3 +148,16 @@ def allow_nan_array(attributes):
             ret.append(entry)
 
     return ret
+
+
+def validate_num_samples_num_iteration(num_rows_subsample, num_iterations):
```
lajohn4747, line 153: Putting this in a separate function as we reuse this for DCROverfittingProtection.

sdmetrics/single_table/privacy/dcr_baseline_protection.py
```diff
@@ -0,0 +1,201 @@
...
+        random_data = cls._generate_random_data(sanitized_real_data)
+
+        sum_synthetic_median = 0
+        sum_random_median = 0
+        sum_score = 0
+
+        for _ in range(num_iterations):
+            synthetic_sample = sanitized_synthetic_data
+            if num_rows_subsample is not None:
+                synthetic_sample = sanitized_synthetic_data.sample(n=num_rows_subsample)
```
R-Palazzo, line 96: We might consider taking the minimum of `num_rows_subsample` and `len(sanitized_synthetic_data)` to avoid row duplication.

sdmetrics/single_table/privacy/dcr_baseline_protection.py
```diff
@@ -0,0 +1,201 @@
...
+
+        Args:
+            real_data (pd.DataFrame):
+                A pd.DataFrame object containing the real data used for training the synthesizer.
+            synthetic_data (pd.DataFrame):
+                A pandas.DataFrame object containing the synthetic data sampled
+                from the synthesizer.
+            real_validation_data (pd.DataFrame):
+                A pandas.DataFrame object containing a holdout set of real data.
+                This data should not have been used to train the synthesizer.
```
R-Palazzo, line 136: Remove this

sdmetrics/single_table/privacy/dcr_baseline_protection.py
```diff
@@ -0,0 +1,201 @@
...
+            metadata,
+            num_rows_subsample,
+            num_iterations,
+        )
+
+        return result.get('score')
+
+    @classmethod
+    def _generate_random_data(cls, real_data):
+        random_data = {}
```
R-Palazzo, line 163: Should we add a seed here or somewhere in the logic?
The main reason is that if I compute the metric using the same real and synthetic data, should I expect to get the same score (when we're not subsampling)? That could be beneficial to test as an integration test as well.

tests/unit/single_table/privacy/test_dcr_baseline_protection.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # Run
+        random_data = DCRBaselineProtection._generate_random_data(real_data)
+
+        # Assert
+        pd.testing.assert_frame_equal(real_data, random_data)
+
+    def test_generate_random_data_different(self):
+        """Test that generated data is differente everytime."""
+        real_data = pd.DataFrame({'float_col': [1.0, 1000.0, 500.0], 'cat_col': ['A', 'B', 'C']})
```
R-Palazzo, line 155: Add # Setup

tests/unit/single_table/privacy/test_dcr_baseline_protection.py
```diff
@@ -0,0 +1,163 @@
...
+        random_data = DCRBaselineProtection._generate_random_data(real_data)
+
+        # Assert
+        for col_name, col_data in random_data.items():
+            if real_data[col_name].isna().any():
+                assert col_data.isna().any()
+
+    def test__generate_random_data_single_value(self):
+        """Test that random generated data for a single value should be the original data."""
+        # Setip
```
R-Palazzo, line 142: `# Setup`

R-Palazzo (commented):
Looking good!

I just let 1 or 2 suggestions

sdmetrics/single_table/privacy/dcr_baseline_protection.py
frances-h: We should probably check that `num_rows_subsample` is smaller than the synthetic data anways, and warn the user and fallback to not subsampling at all. I think the `replace` method defaults to `replace=False` anyways, so currently if `num_rows_subsample` is larger than the length of the synthetic data I would expect this to error.

lajohn4747: Yea similar to how I did DCROverfittingMetric, I will just clear those arguments if `num_rows_subsample` > `len(synthetic_data)`

sdmetrics/single_table/privacy/dcr_baseline_protection.py
```diff
@@ -0,0 +1,201 @@
...
+
+            elif is_datetime(real_data[col]):
+                min_date, max_date = real_data[col].min(), real_data[col].max()
+                total_seconds = (max_date - min_date).total_seconds()
+                random_values = min_date + pd.to_timedelta(
+                    np.random.uniform(0, total_seconds, num_samples), unit='s'
+                )
+
+            else:
+                random_values = real_data[col].sample(num_samples, replace=True).values
```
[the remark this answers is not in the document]
lajohn4747: Yeah I can just default to random values, I think I already covered all the values we would check anyway. 

I saw like PyArrow types, and wasn't sure how to randomly generate so just decided to get different values, but I realize now that is still an experimental thing.

sdmetrics/single_table/privacy/dcr_baseline_protection.py
lajohn4747: For tests I mock the return value of this function.

For our use case, @npatki said it probably shouldn't affect our values too much and if it does we can go back and update the api to pass in the seed.

R-Palazzo (approved):

frances-h (approved):

PR #729, opened 2025-02-21
Use metadata to validate inputs to `BinaryClassifierEfficacy` precision and recall metrics

### LINKED ISSUE #727: Use metadata to validate inputs to `BinaryClassifierEfficacy` precision and recall metrics
### Problem Description

Currently, if a datetime column is passed to either of the `BinaryClassifierEfficacy` metrics but it is not already converted to a `pd.datetime`, the metric will error, even if the metadata provides us with a datetime format. In addition, the metric does not filter out columns not present in the metadata and does not warn if columns specified in the metadata are missing from the data.

### Expected behavior
We should update the metrics so that we convert datetime columns to timestamps. If a datetime format is provided in the metadata, we should use it to convert the column. If a datetime format is not provided, we should still try to convert but show a warning that no datetime format was provided for that column.

Additionally, when data is passed to the metric we should remove any columns in the data that are not in the metadata. If extra columns are passed, we should show a warning and drop the columns before proceeding.
`UserWarning: Some columns ('age', 'salary', 'capitals-gains') are not in the metadata. These columns will not be included in this metric's computation.`

Alternatively, if the data is missing columns that are specified in the metadata, we should also show a warning:
`UserWarning: Some columns ('age', 'salary', 'capital-gains') are in the metadata but they are not present in the data.`

R-Palazzo:
CU-86b3zyra8
Resolve #727

This PR contains 3 main contributions:
- Resolve #727
- Filter out PII columns for the new ML efficacy metrics (Row 3 of the bug hunt sheet but without issue filled)
- Make the `metadata validation` consistent among the entire library as we discussed @amontanez24 

Since news metrics include metadata as a parameter, it's beneficial to always refer to what's defined in _utils_metadata.py to ensure consistency.

### DIFF
--- a/sdmetrics/_utils_metadata.py
+++ b/sdmetrics/_utils_metadata.py
@@ -73,16 +73,28 @@ def handle_single_and_multi_table(single_table_func):
 def _convert_datetime_columns(data, metadata):
     """Convert datetime columns to datetime type."""
     columns_missing_datetime_format = []
+    message_conversion_failed = (
+        'Conversion to datetime failed for the following columns with errors: \n {}'
+    )
+    columns_with_conversion_issues = []
     for column in metadata['columns']:
         if metadata['columns'][column]['sdtype'] == 'datetime':
             is_datetime = pd.api.types.is_datetime64_any_dtype(data[column])
             if not is_datetime:
                 datetime_format = metadata['columns'][column].get('format')
-                if datetime_format:
-                    data[column] = pd.to_datetime(data[column], format=datetime_format)
-                else:
-                    columns_missing_datetime_format.append(column)
-                    data[column] = pd.to_datetime(data[column])
+                try:
+                    if datetime_format:
+                        data[column] = pd.to_datetime(data[column], format=datetime_format)
+                    else:
+                        columns_missing_datetime_format.append(column)
+                        data[column] = pd.to_datetime(data[column])
+                except Exception as e:
+                    columns_with_conversion_issues.append(f"'{column}': {str(e)}")
+
+    if columns_with_conversion_issues:
+        raise ValueError(
+            message_conversion_failed.format('\n'.join(columns_with_conversion_issues))
+        )
 
     if columns_missing_datetime_format:
         columns_to_print = "', '".join(columns_missing_datetime_format)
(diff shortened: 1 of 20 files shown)


sdmetrics/_utils_metadata.py
```diff
@@ -0,0 +1,158 @@
...
+def _validate_multi_table_metadata(metadata):
+    """Validate the metadata for multiple tables."""
+    _validate_metadata_dict(metadata)
+    if 'tables' not in metadata:
+        raise ValueError(
+            "Multi-table metadata must include a 'tables' key that maps table names"
+            ' to their respective metadata.'
+        )
+    for table_name, table_metadata in metadata['tables'].items():
+        _validate_single_table_metadata(table_metadata)
```
frances-h, line 37: Should we update this error message to say which table is invalid?

frances-h (commented):
Looks good, thanks for centralizing this!

sdmetrics/_utils_metadata.py
```diff
@@ -0,0 +1,158 @@
...
+    for column in metadata['columns']:
+        if metadata['columns'][column]['sdtype'] == 'datetime':
+            is_datetime = pd.api.types.is_datetime64_any_dtype(data[column])
+            if not is_datetime:
+                datetime_format = metadata['columns'][column].get('format')
+                if datetime_format:
+                    data[column] = pd.to_datetime(data[column], format=datetime_format)
+                else:
+                    columns_missing_datetime_format.append(column)
+                    data[column] = pd.to_datetime(data[column])
```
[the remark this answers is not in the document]
R-Palazzo: For me it makes sense to crash here because otherwise we will crash later when computing the metric but maybe we can have our own error message

sdmetrics/_utils_metadata.py
```diff
@@ -73,16 +73,28 @@ def wrapper(data, metadata):
...
+                    else:
+                        columns_missing_datetime_format.append(column)
+                        data[column] = pd.to_datetime(data[column])
+                except Exception as e:
+                    columns_with_conversion_issues.append(f"'{column}': {str(e)}")
+
+    if columns_with_conversion_issues:
+        raise ValueError(
+            message_conversion_failed.format('\n'.join(columns_with_conversion_issues))
+        )
```
amontanez24, line 97: can we use [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining) here?

amontanez24 (approved):

frances-h (approved):

PR #733, opened 2025-02-22
Add DCROverfittingProtection metric

### LINKED ISSUE #719: Add `DCROverfittingProtection` metric
### Problem Description
Currently in SDMetrics, we are offering [DisclosureProtection](https://docs.sdv.dev/sdmetrics/metrics/metrics-glossary/disclosureprotection) as the only metric for measuring data privacy. We've vetted this metric to make sure that it's understandable and insightful. However, running this metric requires the user to make some decisions about what to input as known columns and sensitive columns. This means that the privacy score is only ever based on a subset of columns.

We'd like to create a new metric that could be run on an entire set of columns.

### Expected behavior
Add the `DCROverfittingProtection` metric. This metric uses a DCR (distance to closest record) computation to measure whether the synthetic data has been overfit to the real data – as compared to a holdout set.

```python
from sdmetrics.single_table import DCROverfittingProtection

DCROverfittingProtection.compute(
  real_training_data=real_df,
  synthetic_data=synthetic_df,
  real_validation_data=holdout_df,
  metadata=my_metadata)

DCROverfittingProtection.compute_breakdown(
  real_training_data=real_df,
  synthetic_data=synthetic_df,
  real_validation_data=holdout_df,
  metadata=my_metadata)
```

#### `compute_breakdown`
__Args__
- `real_training_data (pd.DataFrame)`: A pandas.DataFrame object containing the real data used for training the synthesizer.
- `synthetic_data (pd.DataFrame)`: A pandas.DataFrame object containing the synthetic data sampled from the synthesizer.
- `real_validation_data (pd.DataFrame)`: A pandas.DataFrame object containing a holdout set of real data. This data should not have been used to train the synthesizer. For this metric to work effectively, this metric should be roughly the same size as the real data.
- `metadata (dict)`: A metadata dictionary that describes the table of data.
- `num_rows_subsample (int, optional)`: The number of synthetic data rows to subsample from the synthetic data. Use this to increase the speed of the computation, if your dataset is large. Default is `None` (do not subsample the data).
- `num_iterations (int, optional)`: The number of iterations to perform when subsampling. The final score will be the average of all iterations. Default is 1.

__Returns__
A dictionary breakdown of the score, with the following information:
- The overall score of the metric
- The percent of synthetic rows that were closer to the training set and the percentage of synthetic rows that were closer to the holdout set.
    - If multiple iterations were used, the percentages should be the average across all iterations.

Example
```python
{
  'score': 0.5,
  'synthetic_data_percentages': {
    'closer_to_training': 0.75,
    'closer_to_holdout': 0.25
  }
}
```

__Computation Algorithm__
1. Preprocessing:
  -> Only consider columns that are statistical (sdtype is numerical, datetime, categorical, or boolean).
  -> Convert all datetimes into UnixTimestamps so they can be treated as numerical values for the rest of the metric
  -> If requested by the user, subsample the synthetic data
2. Go through every row, `s` in the synthetic data and measure the following:
  a. Measure the DCR between `s` and the real data using the algorithm below. Call this distance `d(s, real)`
  b. Measure the DCR of `s`  in the holdout set using the same algorithm. Call this distance `d(s, holdout)`
  -> After Step 1 there will be a `d(s, real)` and `d(s, holdout)` for every row in the synthetic data. 
3. Calculate the percentage of rows where `d(s, real) < d(s, holdout)` (i.e the number of rows that are closer to the real dataset than the holdout dataset). Call this percentage `p`.
4. Return  `min((1-p)*2, 1)`
5. Repeat the above for `num_iterations` requested by the user

_DCR between a row and a dataset_
- Given a row `s` and a dataset `D`, this algorithm computes the distance to the closest record of `D`.
- Compute the record distance between `s` and every single row in `D` using the DCR between two rows algorithm below.
- The DCR is the minimum distance value that we compute between `s` and all rows of `D`.

_DCR between 2 rows (including null values)_
Given a row `s` and a row `d` compute distance between these rows.
- Loop through every column's value (j) in these rows and compute the distance. If these values are called `s_j` and `d_j`, then distance between them is defined based on the type of data:
  - For numerical data, `distance(s_j, d_j) = abs(s_j - d_j) / (range of column)` if both `s_j` and `d_j` are both non-null.
    - The `distance(s_j, d_j) = 0` if both `s_j` and `d_j` are null.
    - The `distance(s_j, d_j) = 1` if only one of `s_j` and `d_j` are null.
  - For categorical data, nulls are considered another category and treated like any other category value.
    - `distance(s_j, d_j) = 0 if s_j == d_j else 1`

The overall distance between rows s and d is the average of the column distances we compute

#### `compute`

The `compute` method should just return the overall `score` parameter calculated by `compute_breakdown`.

### Additional context
See [doc](https://docs.google.com/document/d/1APIEvLfCCptC1bbFgF8bBRLnyAC_3y5pICb_9yCUxuA/edit?tab=t.jkxbunij8p7p#heading=h.va5ntny100zj)

lajohn4747:
resolves #719 
CU-86b3w7pbq


sdv-team:
Task linked: [CU-86b3w7pbq SDMetrics - Add `DCROverfittingProtection` metric #719](https://app.clickup.com/t/86b3w7pbq)

sdmetrics/single_table/privacy/dcr_overfitting_protection.py
```diff
@@ -0,0 +1,204 @@
...
+from sdmetrics.single_table.base import SingleTableMetric
+from sdmetrics.single_table.privacy.dcr_utils import calculate_dcr
+from sdmetrics.utils import get_columns_from_metadata, get_type_from_column_meta
+
+
+class DCROverfittingProtection(SingleTableMetric):
+    """DCR Overfitting Protection metric."""
+
+    name = 'DCROverfittingProtection'
+    goal = Goal.IGNORE
```
frances-h, line 15: We don't actually really use this, but generally we set this to `Goal.MAXIMIZE`

sdmetrics/single_table/privacy/dcr_overfitting_protection.py
```diff
@@ -0,0 +1,204 @@
...
+        num_rows_subsample,
+        num_iterations,
+    ):
+        if num_rows_subsample is not None:
+            if not isinstance(num_rows_subsample, int):
+                raise ValueError('num_rows_subsample must be an integer.')
+            if num_rows_subsample < 1:
+                raise ValueError(
+                    f'num_rows_subsample ({num_rows_subsample}) must be greater than 1.'
+                )
```
frances-h, line 35: You could combine these into one check with 
`if not isinstance(num_rows_subsample, int) or num_rows_subsample < 1:`

Then the error would just be 'num_rows_subsample must be an integer greater than 1.'

sdmetrics/single_table/privacy/dcr_overfitting_protection.py
```diff
@@ -0,0 +1,204 @@
...
+        elif num_rows_subsample is None and num_iterations > 1:
+            raise ValueError(
+                'num_iterations should not be greater than 1 if there is not subsampling.'
+            )
+
+        if not isinstance(num_iterations, int):
+            raise ValueError('num_iterations must be an integer.')
+
+        if num_iterations < 1:
+            raise ValueError(f'num_iterations ({num_iterations}) must be greater than 1.')
```
frances-h, line 45: Same suggestion as the `num_rows_subsample` check

sdmetrics/single_table/privacy/dcr_overfitting_protection.py
lajohn4747: Sounds good, yeah it's still checked in the test `test_single_table.py::test_compute_all`

sdmetrics/single_table/privacy/util.py
```diff
@@ -148,3 +148,16 @@ def allow_nan_array(attributes):
             ret.append(entry)
 
     return ret
+
+
+def validate_num_samples_num_iteration(num_rows_subsample, num_iterations):
```
lajohn4747, line 153: Made this a util function as we will reuse this for DCRBaselineProtection.

tests/integration/single_table/privacy/test_dcr_overfitting_protection.py
```diff
@@ -0,0 +1,138 @@
+import random
+
+import pandas as pd
+from sklearn.model_selection import train_test_split
+
+from sdmetrics.demos import load_single_table_demo
+from sdmetrics.single_table.privacy.dcr_overfitting_protection import DCROverfittingProtection
```
R-Palazzo, line 7: Can we check the import from `sdmetrics.single_table.privacy` here?

R-Palazzo (approved):
LGTM!

sdmetrics/single_table/privacy/dcr_overfitting_protection.py
```diff
@@ -0,0 +1,177 @@
...
+            dcr_holdout = calculate_dcr(validation_data, synthetic_sample, metadata)
+
+            num_rows_closer_to_real = np.where(dcr_real < dcr_holdout, 1.0, 0.0).sum()
+            total_rows = dcr_real.size
+            percentage_close_to_real = num_rows_closer_to_real / total_rows
+            percentage_close_to_random = 1 - percentage_close_to_real
+            score = min((1.0 - percentage_close_to_real) * 2, 1.0)
+            sum_of_scores += score
+            sum_of_p_close_to_real += percentage_close_to_real
+            sum_of_p_close_to_random += percentage_close_to_random
```
frances-h, line 121: Can we be consistent with naming and rename this to `sum_percent_close_to_real`?

sdmetrics/single_table/privacy/dcr_overfitting_protection.py
```diff
@@ -0,0 +1,177 @@
...
+
+        training_data, sanitized_synthetic_data, validation_data = sanitized_data
+
+        sum_of_scores = 0
+        sum_of_p_close_to_real = 0
+        sum_of_p_close_to_random = 0
+        for _ in range(num_iterations):
+            synthetic_sample = sanitized_synthetic_data
+            if num_rows_subsample is not None:
+                synthetic_sample = sanitized_synthetic_data.sample(n=num_rows_subsample)
```
frances-h, line 109: What happens if `num_rows_subsample` is larger than our synthetic data?

lajohn4747: I'll add a check for that too so it fails fast.

frances-h: Instead of erroring out can we warn the user that the data is smaller than the desired subsample and just continue without subsampling?

lajohn4747: Yep that makes sense! I added tests for that case too.

frances-h (approved):
Looks good, thanks for addressing!

PR #736, opened 2025-02-28
Upgrade `pomegranate` dependency to Python 3.13

### LINKED ISSUE #717: Upgrade `pomegranate` dependency to Python 3.13
### Problem Description
In #645, we will add SDMetrics support for Python 3.13. However, this plan does not include the optional, `pomegranate` dependency because of blocking changes (in both API and how the license is shown).

### Expected behavior
[Blocker] We've requested that the pomegranate library update the way their license is shown in [pomegranate #1127](https://github.com/jmschrei/pomegranate/issues/1127).

Once this is done, we can allow our own pomegranate dependency to 3.13 as well. Until then, people who want to use pomegranate alongside SDMetrics will have to stick to Python 3.12 or below.

```
pip install sdmetrics[pomegranate]
```

R-Palazzo:
CU-86b3vfzp4
Resolve #717

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ torch = [
     "torch>=2.2.0;python_version>='3.12' and python_version<'3.13'",
     "torch>=2.6.0;python_version>='3.13'",
 ]
-pomegranate = ['pomegranate>=1.0,<2.0']
+pomegranate = ['pomegranate>=1.1.1,<2.0']
 xgboost = ['xgboost>=2.1.3']
 test = [
     'sdmetrics[torch]',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ torch = [
     "torch>=2.2.0;python_version>='3.12' and python_version<'3.13'",
     "torch>=2.6.0;python_version>='3.13'",
 ]
-pomegranate = ['pomegranate>=1.1.1,<2.0']
+pomegranate = ['pomegranate>=1.1.2,<2.0']
 xgboost = ['xgboost>=2.1.3']
 test = [
     'sdmetrics[torch]',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,9 +63,7 @@ torch = [
 pomegranate = ['pomegranate>=1.1.2,<2.0']
 xgboost = ['xgboost>=2.1.3']
 test = [
-    'xgboost>=2.1.3',
-    'pomegranate>=1.1.2,<2.0',
-    'sdmetrics[torch]',
+    'sdmetrics[pomegranate,torch,xgboost]',
     'pytest>=6.2.5,<7',
     'pytest-cov>=2.6.0,<3',
     'pytest-rerunfailures>=10.3,<15',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ requires-python = ">=3.8,<3.14"
 readme = 'README.md'
 dependencies = [
     "numpy>=1.22.2;python_version<'3.10'",
-    "numpy>=1.23.3;python_version>='3.10' and python_version<'3.12'",
+    "numpy>=1.24.0;python_version>='3.10' and python_version<'3.12'",
     "numpy>=1.26.0;python_version>='3.12' and python_version<'3.13'",
     "numpy>=2.1.0;python_version>='3.13'",
     "pandas>=1.4.0;python_version<'3.11'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ torch = [
     "torch>=2.2.0;python_version>='3.12' and python_version<'3.13'",
     "torch>=2.6.0;python_version>='3.13'",
 ]
-pomegranate = ['pomegranate>=0.15,<1.0']
+pomegranate = ['pomegranate>=1.0,<2.0']
 xgboost = ['xgboost>=2.1.3']
 test = [
     'sdmetrics[torch]',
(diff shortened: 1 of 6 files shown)


sdv-team:
Task linked: [CU-86b3vfzp4 SDMetrics - Upgrade `pomegranate` dependency to Python 3.13 #717](https://app.clickup.com/t/86b3vfzp4)

tests/unit/single_table/test_bayesian_network.py
```diff
@@ -1,43 +1,68 @@
...
 
-        # Act and Assert
-        expected_message = r'Please install pomegranate with `pip install sdmetrics\[pomegranate\]`\. Python 3\.13 is not supported\.'  # noqa: E501
-        with pytest.raises(ImportError, match=expected_message):
-            metric.compute(Mock(), Mock())
+        # Run
+        result = metric.compute(real_data, synthetic_data, metadata)
+
+        # Assert
+        assert np.isclose(result, 0.1111111044883728, atol=1e-5)
```
R-Palazzo, line 54: Those tests ensure the metric computation remains consistent while upgrading pomegranate.

pyproject.toml
```diff
@@ -60,7 +60,7 @@ torch = [
     "torch>=2.2.0;python_version>='3.12' and python_version<'3.13'",
     "torch>=2.6.0;python_version>='3.13'",
 ]
-pomegranate = ['pomegranate>=0.15,<1.0']
+pomegranate = ['pomegranate>=1.1.1,<2.0']
```
rwedge, line 63: We should make 1.1.2 our minimum dependency to ensure package metadata has the changes we asked for

sdmetrics/single_table/bayesian_network.py
```diff
@@ -16,14 +18,6 @@ class BNLikelihoodBase(SingleTableMetric):
 
     @classmethod
     def _likelihoods(cls, real_data, synthetic_data, metadata=None, structure=None):
-        try:
-            from pomegranate import BayesianNetwork
-        except ImportError:
-            raise ImportError(
-                'Please install pomegranate with `pip install sdmetrics[pomegranate]`. '
-                'Python 3.13 is not supported.'
```
rwedge, line 24: I think we still want this import error but we can drop the Python 3.13 sentence

pyproject.toml
```diff
@@ -60,7 +60,7 @@ torch = [
     "torch>=2.2.0;python_version>='3.12' and python_version<'3.13'",
     "torch>=2.6.0;python_version>='3.13'",
 ]
-pomegranate = ['pomegranate>=0.15,<1.0']
+pomegranate = ['pomegranate>=1.1.2,<2.0']
 xgboost = ['xgboost>=2.1.3']
 test = [
     'sdmetrics[torch]',
```
rwedge, line 66: Since the tests won't run if pomegranate and xgboost aren't installed, let's add them to the 'test' optional dependency group.  Then we can also update the github workflows to install `.[test]` instead of `.[pomegranate,xgboost,test]`

R-Palazzo: Yes that's a good point, done in [9d9dc59](https://github.com/sdv-dev/SDMetrics/pull/736/commits/9d9dc59f27c26e7b458a2388fa53ccb94873398b)

pyproject.toml
```diff
@@ -63,6 +63,8 @@ torch = [
 pomegranate = ['pomegranate>=1.1.2,<2.0']
 xgboost = ['xgboost>=2.1.3']
 test = [
+    'xgboost>=2.1.3',
+    'pomegranate>=1.1.2,<2.0',
     'sdmetrics[torch]',
```
rwedge, line 68:
```suggestion
    'sdmetrics[pomegranate,torch,xgboost]',
```

sdmetrics/single_table/bayesian_network.py
```diff
@@ -34,19 +34,28 @@ def _likelihoods(cls, real_data, synthetic_data, metadata=None, structure=None):
...
+                    pd.concat([real_data[column], synthetic_data[column]]).unique()
+                )
+            }
+            for column in fields
+        }
+        for column in fields:
+            real_data[column] = real_data[column].map(category_to_integer[column]).astype('int64')
+            synthetic_data[column] = (
+                synthetic_data[column].map(category_to_integer[column]).astype(int)
+            )
```
amontanez24, line 51: why do we need to do this conversion? Also, can't we do this without a for loop by just doing something like real_data[fields] = ...

amontanez24: Why didn't we need the conversion before?

R-Palazzo: They introduced a `bn.fit()` method that expects numerical values: https://pomegranate.readthedocs.io/en/latest/api.html. 

I believe that in the previous version on Cython, they handled some conversions themselves.

amontanez24: Is our method of conversion the same as theirs was? I just want to make sure the behavior is the same

R-Palazzo: Yes I think, I checked their code previous 1.0 and there have those lines:
https://github.com/jmschrei/pomegranate/blob/9400c9c04808a5dd5a72da6f010519328d42de6b/pomegranate/BayesianNetwork.pyx#L1512
https://github.com/jmschrei/pomegranate/blob/9400c9c04808a5dd5a72da6f010519328d42de6b/pomegranate/BayesianNetwork.pyx#L1539

Also feel free to run the unit test I added on an old pomegranate version (before 1.0); the score should be the same or very close to the current one.

rwedge (approved):

amontanez24 (approved):

PR #745, opened 2025-03-14
DCR calculation, handle if numerical/datetime range is 0

### LINKED ISSUE #739: DCR metrics should allow the possibility of constant numerical columns
### Problem Description
The DCRBaselineProtection and DCROverfittingProtection metrics measure the privacy of synthetic data by measuring how far away it is from the real data. Underlying both of these metrics is a computation that measures the "distance" between 2 rows of data. It is defined in issue #720 (pasted below):

Image

**The problem**: Numerical data may sometimes have a range of 0. This means that in the real data, all values in the column were exactly the same. 

Currently, both DCR metric would crash on such a column because `col_range=0` leads to a `ZeroDivisionError`.

### Expected behavior
If, for numerical data column, the real column range = 0, then compute the distance using the same methodology we do for categorical column. That is to say: `distance(s_j, d_j) = 0` if `s_j == d_j` else the distance is `1`

lajohn4747:
resolves #739 
CU-86b48qmjv

Handles the case where numerical or datetime range == 0 for `sdmetrics/single_table/privacy/dcr_utils.py`.
Will handle the other divide by zero errors in separate PRs.

### DIFF
--- a/sdmetrics/single_table/privacy/dcr_utils.py
+++ b/sdmetrics/single_table/privacy/dcr_utils.py
@@ -40,7 +40,9 @@ def _calculate_dcr_value(synthetic_value, real_value, sdtype, col_range=None):
         if isinstance(difference, pd.Timedelta):
             difference = difference.total_seconds()
 
-        distance = difference / col_range
+        distance = 0.0 if synthetic_value == real_value else 1.0
+        if col_range != 0:
+            distance = difference / col_range
         return min(distance, 1.0)
 
     if synthetic_value == real_value:
--- a/tests/unit/single_table/privacy/test_dcr_utils.py
+++ b/tests/unit/single_table/privacy/test_dcr_utils.py
@@ -333,8 +333,44 @@ def test_calculate_dcr_with_shuffled_data():
     metadata = {'columns': {'num_col': {'sdtype': 'numerical'}}}
 
     # Run
-    result = calculate_dcr(synthetic_df, real_df, metadata)
-    result_shuffled = calculate_dcr(synthetic_df_shuffled, real_df_shuffled, metadata)
+    result = calculate_dcr(synthetic_data=synthetic_df, real_data=real_df, metadata=metadata)
+    result_shuffled = calculate_dcr(
+        synthetic_data=synthetic_df_shuffled, real_data=real_df_shuffled, metadata=metadata
+    )
 
     # Assert
     check_if_value_in_threshold(result.sum(), result_shuffled.sum(), 0.000001)
+
+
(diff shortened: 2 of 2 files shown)


sdv-team:
Task linked: [CU-86b48qmjv SDMetrics - DCR metrics should allow the possibility of constant numerical columns #739](https://app.clickup.com/t/86b48qmjv)

tests/unit/single_table/privacy/test_dcr_utils.py
```diff
@@ -333,8 +333,44 @@ def test_calculate_dcr_with_shuffled_data():
...
+    result = calculate_dcr(synthetic_data=synthetic_df, real_data=real_df, metadata=metadata)
+    result_shuffled = calculate_dcr(
+        synthetic_data=synthetic_df_shuffled, real_data=real_df_shuffled, metadata=metadata
+    )
 
     # Assert
     check_if_value_in_threshold(result.sum(), result_shuffled.sum(), 0.000001)
+
+
+@pytest.mark.filterwarnings('error')
```
lajohn4747, line 345: I learned that numpy types doesn't throw an zero division error but a runtime warning, so this is to ensure that either case will fail in this test.

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -40,7 +40,9 @@ def _calculate_dcr_value(synthetic_value, real_value, sdtype, col_range=None):
         if isinstance(difference, pd.Timedelta):
             difference = difference.total_seconds()
 
-        distance = difference / col_range
+        distance = 0.0 if synthetic_value == real_value else 1.0
+        if col_range != 0:
+            distance = difference / col_range
```
R-Palazzo, line 45: Add a blank line

tests/unit/single_table/privacy/test_dcr_utils.py
```diff
@@ -333,8 +333,44 @@ def test_calculate_dcr_with_shuffled_data():
...
+        'num_col': synthetic_data_num_same,
+        'date_col': synthetic_data_date_same,
+    })
+    synthetic_df_half = pd.DataFrame({
+        'num_col': synthetic_data_num_diff,
+        'date_col': synthetic_data_date_same,
+    })
+    metadata = {'columns': {'num_col': {'sdtype': 'numerical'}, 'date_col': {'sdtype': 'datetime'}}}
+
+    # Run
```
R-Palazzo, line 370: `# Run and Assert`

frances-h (approved):

PR #746, opened 2025-03-17
Dcr baseline fix calculation

### LINKED ISSUE #738: The DCRBaselineProtection metric crashes when the distance between random data and real data is 0
### Environment Details

- SDMetrics version: 0.19.1 (DCR Branch)
- Python version: Python 3.11
- Operating System:  Linux Colab

### Error Description
The new `DCRBaselineProtection` metric is a measure of privacy of the synthetic data. It asks the Q: _If I were to use random data instead of synthetic data, how much more private would it be?_

The metric is based on the distance to closest record. It measures:
- `random_data_median`: The typical distance between random data and real
- `synthetic_data_median`: The typical distance between synthetic data and real

The final score is: `synthetic_data_median / random_data_median`

However in some cases, `random_data_median=0`. This happens when you have a dataset that is capable of very little diversity. For eg, the dataset has only 2 columns, which each can only contain 2 possible discrete values (=4 possibilities). 

| is_active | response |
|-----------|----------|
| True      | "YES"    |
| True      | "NO"     |
| False     | YES      |
| False     | "NO"     |

If `random_data_median=0`, this metric currently crashes with a `ZeroDivisionError`.

### Expected Behavior
Rather than crashing, the final metric score should be `NaN`, indicating that it was is not recommended to be computing privacy on such a dataset anyways.

The `compute_breakdown` should still return the individual median scores so the user can understand more about what's happening.

```python
>>> DCRBaselineProtection.compute_breakdown(
  real_data=real_df,
  synthetic_data=synthetic_df,
  metadata=my_metadata)
{
  'score': NaN,
  'median_DCR_to_real_data': {
    'synthetic_data': 0.25
    'random_data_baseline': 0.0
  }
}
`` `

### LINKED ISSUE #742: The `DCRBaselineProtection` metric is not producing the expected score
### Environment Details
* Python version: Python 3.11
* Operating System: Linux Colab

### Error Description
The `DCRBaselineProtection` metric is not producing the correct score. There seems to be something wrong in the computation of the median DCR between a dataset and the real data.

We discussed there potentially being 2 different root causes to this issue:

1. Making sure that the DCR computation done for: `synthetic vs. real` and `random vs. real` datasets. More particularly, we want to compare:
    - The _median_ DCR for a synthetic data row and the real dataset and
    - The _median_ DCR for a random data row and the real dataset
2. For numerical data, the distance computation should be based on the range of the _real data column_ (not the synthetic data or random data)

### Steps to reproduce

In this case, I expect the median distance between synthetic and real to be 0.1 
```python
from sdmetrics.single_table.privacy import DCRBaselineProtection
import pandas as pd
import numpy as np

real_data = pd.DataFrame(data={
    'A': [0, 10, 3, 4, 1]}) # the range of this column is 10

synthetic_data = pd.DataFrame(data={
    'A': [5],}) # the DCR between this row and the real dataset is 1/10 = 0.1 

metadata = {
    'columns': {
        'A': { 'sdtype': 'numerical' },}}

# I expect that the median distance between real data and synthetic is 0.1
DCRBaselineProtection.compute_breakdown(
    real_data=real_data,
    synthetic_data=synthetic_data,
    metadata = metadata)
```

In this case, I expect the median distance between synthetic and real to be 0 because there is a null value in both the real and synthetic data.
```python
from sdmetrics.single_table.privacy import DCRBaselineProtection
import pandas as pd
import numpy as np

real_data = pd.DataFrame(data={
    'A': [0, 10, 3, 4, 1, np.nan], })

synthetic_data = pd.DataFrame(data={
    'A': [np.nan], # the DCR between this an the real data is 0 (np.nan exactly matches)})

metadata = {
    'columns': {
        'A': { 'sdtype': 'numerical' }}}

# I expect that the median distance between real data and synthetic is 0
DCRBaselineProtection.compute_breakdown(
    real_data=real_data,
    synthetic_data=synthetic_data,
    metadata = metadata)
```

### LINKED ISSUE #743: The `DCRBaselineProtection` metric is not creating the correct size of random data
### Environment Details
* Python version: Python 3.11
* Operating System: Linux Colab

### Error Description
The new `DCRBaselineProtection` metric is a measure of privacy of the synthetic data. It asks the Q: _If I were to use random data instead of synthetic data, how much more private would it be?_

For an accurate comparison point, the size of the random data should be the same as the size of the synthetic data. Otherwise, the distance to closest record calculation may give an unfair advantage to either the synthetic or random dataset.

By default, the size of the random dataset is correct. However, if I use the `num_rows_subsample` option, it is not correct. 

For eg, let's say my synthetic data has 50K rows but when calling the metric, I ask for a subsample of 1000 rows only. In this case, the metric should create only 1000 random data rows (to match the synthetic data subsample). Instead, it is currently creating the full 50K rows.

```python
>>> DCRBaselineProtection.compute(
      real_data=real_df,
      synthetic_data=synthetic_df,
      metadata=my_metadata,
      num_rows_subsample=1000,
      num_iterations=3)
0.58808
```

lajohn4747:
resolves #743 
resolves #742 
resolves #738 

CU-86b48qmjr
CU-86b48rkqg
CU-86b48rkqz

The fixes:
- DCRBaselineProtection was comparing synthetic to real and synthetic to random. This should be synthetic vs real, and random vs real
- DCRBaselineProtection was putting the wrong inputs into `calculate_dcr`. Synthetic and real got swapped and thus the range was based on synthetic not real. This was fixed by labeling args
- Size of random data to compare should be the same as the sample size of the synthetic. This PR should update it so that we create a random data from the real data and compare random samples of size equal to the synthetic sample
- If `random_median == 0` we should make the score a `nan` instead of erroring.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -32,16 +32,10 @@ jobs:
       run: |
           brew install libomp
           echo 'export DYLD_LIBRARY_PATH=$(brew --prefix libomp)/lib:$DYLD_LIBRARY_PATH' >> $GITHUB_ENV
-    - if: matrix.python-version != 3.13
-      name: Install dependencies
+    - name: Install dependencies
       run: |
           python -m pip install --upgrade pip
-          python -m pip install invoke .[pomegranate,xgboost,test]
-    - if: matrix.python-version == 3.13
-      name: Install dependencies
-      run: |
-          python -m pip install --upgrade pip
-          python -m pip install invoke .[xgboost,test]
+          python -m pip install invoke .[test]
     - name: Run integration tests
       run: invoke integration
 
--- a/.github/workflows/minimum.yml
+++ b/.github/workflows/minimum.yml
@@ -35,6 +35,6 @@ jobs:
     - name: Install dependencies
       run: |
           python -m pip install --upgrade pip
-          python -m pip install invoke .[test,xgboost]
+          python -m pip install invoke .[test]
     - name: Test with minimum versions
       run: invoke minimum
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Install dependencies
       run: |
           python -m pip install --upgrade pip
-          python -m pip install invoke .[test,xgboost]
+          python -m pip install invoke .[test]
     - name: Run unit tests
       run: invoke unit
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,26 @@
 # History
 
+## v0.19.0 - 2025-02-24
+
+### New Features
+
+* Use metadata to validate inputs to `BinaryClassifierEfficacy` precision and recall metrics - Issue [#727](https://github.com/sdv-dev/SDMetrics/issues/727) by @R-Palazzo
+* Speed up calculation of the QualityReport - Issue [#718](https://github.com/sdv-dev/SDMetrics/issues/718) by @R-Palazzo
+* Allow subsampling when computing the `ContingencySimilarity` metric - Issue [#716](https://github.com/sdv-dev/SDMetrics/issues/716) by @R-Palazzo
+* Add `BinaryClassifierRecallEfficacy` metric - Issue [#712](https://github.com/sdv-dev/SDMetrics/issues/712) by @R-Palazzo
+* Add `BinaryClassifierPrecisionEfficacy` metric - Issue [#711](https://github.com/sdv-dev/SDMetrics/issues/711) by @R-Palazzo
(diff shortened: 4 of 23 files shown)


sdv-team:
Task linked: [CU-86b48qmjr SDMetrics - The DCRBaselineProtection metric crashes when the distance between random data and real data is 0 #738](https://app.clickup.com/t/86b48qmjr)
Task linked: [CU-86b48rkqg SDMetrics - The `DCRBaselineProtection` metric is not creating the correct size of random data #743](https://app.clickup.com/t/86b48rkqg)
Task linked: [CU-86b48rkqz SDMetrics - The `DCRBaselineProtection` metric is not producing the expected score #742](https://app.clickup.com/t/86b48rkqz)

frances-h (approved):

PR #751, opened 2025-03-19
Improve handling of datetime columns for DCR metrics

### LINKED ISSUE #740: Improve handling of datetime columns for metrics
### Problem Description
Many different privacy metrics (including DCRBaselineProtection, DCROverfittingProtection, BinaryClassifierPrecisionEfficacy, etc.) allow the user to input _metadata_ that describes the different column's sdtypes.

The handling of _datetime_ columns is currently done via the internal [`_convert_datetime_columns` function](https://github.com/sdv-dev/SDMetrics/blob/main/sdmetrics/_utils_metadata.py#L73). We should improve this function for a more consistent handling of datetime columns.

### Expected behavior

**SDMetrics should not be responsible for trying to infer datetime formats from datetime strings.** This means that SDMetrics should only accept the following types of inputs:
1. Columns that are present as `string` values _with accompanying metadata_ that includes a `datetime_format` string
2. Columns that are already converted to `pd.datetime` (In this case no `datetime_format` specification is needed)

If the datetime values are present as `string` but there is no `datetime_format` in the metadata, we should throw an error. 
```
Error: Datetime column 'start_date' does not have a specified 'datetime_format'. Please add a the required 
datetime_format to the metadata or convert this column to 'pd.datetime' to bypass this requirement.
```

**SDMetrics should expect metadata in the SDV 1.0 format.** This means that the datetime format should be specified via the `'datetime_format'` property ... _not `'format'`_.

lajohn4747:
resolves #740 
CU-86b48rad5

This PR addresses these issues:
- `_convert_datetime_columns` should look for the new format key in metadata called `datetime_format`
- `_convert_datetime_columns` should avoid trying to infer datetime format for strings if is not given in the metadata
- `_generate_random_data` should take into account the granularity of the time data
- `_generate_random_data` uses the correct sanitized real data


sdv-team:
Task linked: [CU-86b48rad5 SDMetrics - Improve handling of datetime columns for metrics #740](https://app.clickup.com/t/86b48rad5)

frances-h (approved):

sdmetrics/_utils_metadata.py
```diff
@@ -72,31 +72,18 @@ def wrapper(data, metadata):
...
-    if columns_missing_datetime_format:
-        columns_to_print = "', '".join(columns_missing_datetime_format)
-        warnings.warn(
-            f'No `datetime_format` provided in the metadata when trying to convert the columns'
-            f" '{columns_to_print}' to datetime. The format will be inferred, but it may not"
-            ' be accurate.',
-            UserWarning,
-        )
+                        f"No 'datetime_format' was described in metadata for '{column}'. "
+                        'Cannot convert objects into datetime formats.'
```
R-Palazzo, line 85: Should we use the error message of the issue here:
```
Error: Datetime column 'start_date' does not have a specified 'datetime_format'. Please add a the required 
datetime_format to the metadata or convert this column to 'pd.datetime' to bypass this requirement.
```

Or maybe just update the current one to explain the two options (Be a `pd.datetime` column or have a `datetime_format` inside the metadata)

R-Palazzo (commented):
This looks good!

I just have one or two small questions

sdmetrics/_utils_metadata.py
lajohn4747: Updated to use the error message

R-Palazzo (approved):

PR #752, opened 2025-03-20
Add private variable seed for the generate data function in DCRBaseline

### LINKED ISSUE #744: Allow me to control randomization when using the `DCRBaslineProtection` metric
### Problem Description
The `DCRBaselineProtection` metric measures the privacy of synthetic data by comparing it against random data. The random data is created by uniformly sampling in the real data's space.

Since randomness is involved in this step, the means that the metric can produce different scores even if it is called on the same data. The variation between these scores should not be too high (TBD) but it would be nice to have a way to control the randomness to produce a deterministic score.

### Expected behavior
We can fix a seed internally for the creation of random data. By default, the seed will _not_ be fixed, meaning that the metric can product a different score.

However, there can be a hidden, private attribute to the class that, when set, will fix the seed. This is a hidden, private attribute, so nothing is promised to the end-user yet. But in the future, this will easily allow us to add a parameter for it if determinism is required.

```python
# there is no seed by default
>>> DCRBaselineProtection._seed
None 

# you can set the seed
>>> DCRBaselineProtection._seed = 23

# now every time the metric is called, it will produce the same result
>>> score = DCRBaselineProtection.compute(...)
>>> score = DCRBaselineProtection.compute(...)

# you can unset the seed to go back to randomness
>>> DCRBaselineProtection._seed = None
```

lajohn4747:
resolves #744 
CU-86b48uvpu

Create a generator with the given seed for the `DCRBaselineProtection` method `_generate_random_data`

### DIFF
--- a/tests/integration/single_table/privacy/test_dcr_baseline_protection.py
+++ b/tests/integration/single_table/privacy/test_dcr_baseline_protection.py
@@ -239,3 +239,23 @@ class TestDCRBaselineProtection:
         assert result['median_DCR_to_real_data']['synthetic_data'] == 0.2
         if result['median_DCR_to_real_data']['random_data_baseline'] == 0.0:
             assert np.isnan(result['score'])
+
+    def test_end_to_end_with_seed(self):
+        """Test end to end with a simple single synthetic value."""
+        # Setup
+        real_data = pd.DataFrame(data={'A': [2, 10, 3, 4, 1, 0]})
+        synthetic_data = pd.DataFrame(data={'A': [12, 32, 42, 15, 19]})
+        metadata = {'columns': {'A': {'sdtype': 'numerical'}}}
+        DCRBaselineProtection._seed = 123
+
+        # Run
+        result = DCRBaselineProtection.compute_breakdown(
+            real_data=real_data,
+            synthetic_data=synthetic_data,
+            metadata=metadata,
+        )
+
+        # Assert
+        assert result['median_DCR_to_real_data']['synthetic_data'] == 0.9
+        assert result['median_DCR_to_real_data']['random_data_baseline'] == 0.1
+        assert result['score'] == 1.0
--- a/tests/integration/single_table/privacy/test_dcr_baseline_protection.py
+++ b/tests/integration/single_table/privacy/test_dcr_baseline_protection.py
@@ -182,11 +182,9 @@ class TestDCRBaselineProtection:
         )
 
         # Assert
-        print(result)
         assert result['median_DCR_to_real_data']['synthetic_data'] == 0.2
         if result['median_DCR_to_real_data']['random_data_baseline'] == 0.0:
             assert np.isnan(result['score'])
-        assert False
 
     def test_end_to_end_with_datetimes(self):
         """Test end to end with a single datetime synthetic value."""
--- a/tests/integration/single_table/privacy/test_dcr_baseline_protection.py
+++ b/tests/integration/single_table/privacy/test_dcr_baseline_protection.py
@@ -217,8 +217,9 @@ class TestDCRBaselineProtection:
 
         # Run
         error_fmt_msg = (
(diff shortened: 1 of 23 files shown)


sdv-team:
Task linked: [CU-86b48uvpu SDMetrics - Allow me to control randomization when using the `DCRBaselineProtection` metric #744](https://app.clickup.com/t/86b48uvpu)

frances-h (approved):

tests/unit/single_table/privacy/test_dcr_baseline_protection.py
```diff
@@ -169,6 +169,19 @@ def test_generate_random_data_different(self):
         with pytest.raises(AssertionError):
             pd.testing.assert_frame_equal(random_data_1, random_data_2)
 
+    def test__generate_random_data_with_seed(self):
+        """Test the generated data with a seed is the same every run."""
+        # Setup
+        real_data = pd.DataFrame({'float_col': [1.0, 1000.0, 500.0], 'cat_col': ['A', 'B', 'C']})
+        DCRBaselineProtection._seed = 5
```
R-Palazzo, line 176: I think it's best to patch the seed here; otherwise, it might set the seed for all the other tests

tests/integration/single_table/privacy/test_dcr_baseline_protection.py
```diff
@@ -239,3 +239,23 @@ def test_end_to_end_with_datetimes(self):
...
+        synthetic_data = pd.DataFrame(data={'A': [12, 32, 42, 15, 19]})
+        metadata = {'columns': {'A': {'sdtype': 'numerical'}}}
+        DCRBaselineProtection._seed = 123
+
+        # Run
+        result = DCRBaselineProtection.compute_breakdown(
+            real_data=real_data,
+            synthetic_data=synthetic_data,
+            metadata=metadata,
+        )
```
R-Palazzo, line 256: Just as a sanity check, if we rerun the same command, we're expecting the exact same result now?
Could we add some checks here that this is the case when the seed is set and not the case when the seed is `None`

R-Palazzo (commented):
This looks good!

I just let 1-2 suggestions.

R-Palazzo (approved):
LGTM!
Thanks for addressing the comments 👍

yangysc:
Thanks for the fix in #752. I think it addresses the main reproducibility issue mentioned in #744, but there are still two stochastic paths that are not controlled by `_seed`.

1. Missing-value sampling still uses the global NumPy RNG:

```python
nan_mask = np.random.rand(num_samples) < nan_ratio
```

So when `real_data` contains NaNs, repeated calls with the same `_seed` may still generate different random baselines.

I suggest changing it to:

```python
nan_mask = randomizer.random(num_samples) < nan_ratio
```

2. Subsampling in `compute_breakdown` is also not controlled by `_seed`:

```python
synthetic_data.sample(n=num_rows_subsample)
random_data.sample(n=num_rows_subsample)
real_data.sample(n=num_rows_subsample)
```

Therefore, when `num_rows_subsample` is enabled, repeated calls with the same `_seed` can still produce different scores.

A possible fix is to create a local RNG from `_seed` and use it for subsampling as well, for example:

```python
randomizer = np.random.default_rng(cls._seed)

synthetic_sample = synthetic_data.sample(
    n=num_rows_subsample,
    random_state=randomizer,
)
random_sample = random_data.sample(
    n=num_rows_subsample,
    random_state=randomizer,
)
real_sample = real_data.sample(
    n=num_rows_subsample,
    random_state=randomizer,
)
```

Alternatively, row indices could be sampled directly with `randomizer.choice()`.

It may also be worth adding regression tests for:

* reproducibility when the real data contains NaNs;
* reproducibility when `num_rows_subsample` is enabled.

So I think #752 fixes the common case, but does not yet make the full metric deterministic under `_seed`.

PR #754, opened 2025-03-24
Optimize DCR calculation using vector operations

lajohn4747:
In order to improve the speed of DCR calculations, we end up cross-merging the datasets (we do this piecemeal of 1000 rows) in order to run vectorized operations instead of running an `apply` to the dataframes.

### DIFF
--- a/sdmetrics/single_table/privacy/dcr_utils.py
+++ b/sdmetrics/single_table/privacy/dcr_utils.py
@@ -1,5 +1,6 @@
 """Distance to closest record measurement functions."""
 
+import numpy as np
 import pandas as pd
 
 from sdmetrics._utils_metadata import _process_data_with_metadata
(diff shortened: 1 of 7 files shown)


sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -1,156 +1,114 @@
...
-                col_range = col_range.total_seconds()
+            if sdtype in ['numerical', 'datetime']:
+                col_range = reference_copy[col_name].max() - reference_copy[col_name].min()
+                if isinstance(col_range, pd.Timedelta):
+                    col_range = col_range.total_seconds()
+                ranges[col_name] = col_range
 
-        column_ranges[col_name] = col_range
+    if not cols_to_keep:
+        raise ValueError('There are no overlapping statistical columns to measure.')
```
frances-h, line 48: Can we add a unit test for this? Looks like this is why codecov is failing

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -1,156 +1,114 @@
...
+        chunk = dataset_copy.iloc[chunk_start : chunk_start + chunk_size].copy()
+        chunk['index'] = range(chunk_start, chunk_start + len(chunk))
+
+        reference_copy['index'] = range(len(reference_copy))
+        full_dataset = chunk.merge(reference_copy, how='cross', suffixes=('_data', '_ref'))
+
+        for col_name in cols_to_keep:
+            sdtype = metadata['columns'][col_name]['sdtype']
+            if sdtype in ['numerical', 'datetime']:
+                diff = (full_dataset[col_name + '_ref'] - full_dataset[col_name + '_data']).abs()
```
frances-h, line 70: Since we're indexing into the datasets a lot, it might be a bit easier to read if we assigned the columns we're working with to `ref_column` and `data_column` and used those variables instead

frances-h (commented):
Looks good, just one suggestion for readability

sdmetrics/single_table/privacy/dcr_utils.py
```diff
@@ -1,156 +1,114 @@
...
+
+    reference_copy = reference_copy[cols_to_keep]
+    reference_copy['index'] = [i for i in range(len(reference_copy))]
+
+    results = []
+    chunk_size = 1000
+
+    for chunk_start in range(0, len(dataset_copy), chunk_size):
+        chunk = dataset_copy.iloc[chunk_start : chunk_start + chunk_size].copy()
+        chunk['index'] = range(chunk_start, chunk_start + len(chunk))
```
R-Palazzo, line 62: Should the `index` column already contain the correct values here based on line 55?
It may be better to use' loc' rather than' iloc' in this case and directly look the actual `index` column.

lajohn4747: Sorry removed this. I think `iloc` is ideal since we are slicing based on row positions

sdmetrics/single_table/privacy/dcr_utils.py
lajohn4747: Strange since `test_calculate_dcr_different_cols_in_metadata` tests this path. Let me rebase with my updates to see if everything is fine.

frances-h (approved):
Minor nitpick, otherwise looks good!

R-Palazzo (approved):
LGTM!

PR #755, opened 2025-03-27
Consolidate handling of datetime columns

### LINKED ISSUE #741: Consolidate handling of datetime columns
### Problem Description
Our reports (Quality Report, Diagnostic Report) and some of our metrics (eg. DCRBaselineProtection, etc.) currently require the user to input metadata, which describes the different sdtypes in each column. Based on the sdtype, the report/metric internally applies different computation to the column.

For datetime columns, the report/metric is also ultimately responsible for converting different formats. But currently, there are two different places we are doing this:

- Metrics are using the [`_convert_datetime_columns`](https://github.com/sdv-dev/SDMetrics/blob/main/sdmetrics/_utils_metadata.py#L73) function
- Report are using the this other [`convert_to_datetime`](https://github.com/sdv-dev/SDMetrics/blob/main/sdmetrics/reports/utils.py#L38) function

We should consolidate this logic into a single utility function.

### Expected behavior
Consolidate the datetime logic into a single utility function. Make sure the logic follows the behavior below.

**SDMetrics should not be responsible for trying to infer datetime formats from datetime strings.** This means that SDMetrics should only accept the following types of inputs:
- Columns that are present as string values with accompanying metadata that includes a `datetime_format` string OR
- Columns that are already converted to `pd.datetime` (In this case no `datetime_format` specification is needed)

If the datetime values are present as string but there is no datetime_format in the metadata, we should throw an error.
```
Error: Datetime column 'start_date' does not have a specified 'datetime_format'. Please add a the required 
datetime_format to the metadata or convert this column to 'pd.datetime' to bypass this requirement.
```

**SDMetrics should expect metadata in the SDV 1.0 format.** This means that the datetime format should be specified via the 'datetime_format' property ... not 'format'.

### Additional Context
In #740 we are updating the logic of the [`_convert_datetime_columns`](https://github.com/sdv-dev/SDMetrics/blob/main/sdmetrics/_utils_metadata.py#L73) function that all the metrics use to the same logic described above.

frances-h:
Resolve #741

### DIFF
--- a/tests/unit/timeseries/test_timeseries.py
+++ b/tests/unit/timeseries/test_timeseries.py
@@ -1,3 +1,5 @@
+import re
+
 import pandas as pd
 import pytest
 
@@ -25,7 +27,8 @@ def test__validate_inputs_for_TimeSeriesMetric():
     }
 
     # Run and Assert
-    with pytest.raises(ValueError, match="Column 'visits' is not a valid datetime"):
+    expected_msg = re.escape("Error converting column 'visits' to timestamp: ")
+    with pytest.raises(ValueError, match=expected_msg):
         TimeSeriesMetric._validate_inputs(
             real_data=df1, synthetic_data=df2, sequence_key=['s_key'], metadata=metadata
         )
--- a/sdmetrics/_utils_metadata.py
+++ b/sdmetrics/_utils_metadata.py
@@ -91,7 +91,9 @@ def _convert_datetime_columns(data, metadata):
     """Convert datetime columns to datetime type."""
     for column in metadata['columns']:
         if metadata['columns'][column]['sdtype'] == 'datetime':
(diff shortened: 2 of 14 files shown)


pvk-developer (approved):
LGTM!

tests/unit/timeseries/test_timeseries.py
```diff
@@ -25,7 +27,8 @@ def test__validate_inputs_for_TimeSeriesMetric():
     }
 
     # Run and Assert
-    with pytest.raises(ValueError, match="Column 'visits' is not a valid datetime"):
+    expected_msg = re.escape("Error converting column 'visits' to timestamp: ")
```
R-Palazzo, line 30: Should the actual error be included in the message?

R-Palazzo (approved):

PR #756, opened 2025-03-27
DCRBaseline and DCROverfitting throws incorrect warnings about missing columns.

### LINKED ISSUE #737: DCROverfitting and DCRBaseline metrics produce too many warnings about missing columns.
### Environment Details

* Python version: Python 3.11
* Operating System: Jupyter, MacOS

### Error Description

User Warning appears multiple times:
```
'The columns ('billing_address', 'credit_card_number', 'guest_email') are in the metadata but they are not present in the data.' 
```
This is caused by the `_process_data_with_metadata` -> `_remove_missing_columns_metadata` multiple times when sanitize the input.

_[Edit from Neha] In our meeting on Mar 13, 2025 we discussed the following as the root cause: Internally, the metrics will drop any columns that are not used for the computation (i.e. id and PII columns). However, these columns are not removed from the metadata, which is causing this consistency. The warnings do not have any impact on the final score._

### Steps to reproduce
From BugHunt [link](https://colab.research.google.com/drive/1vU_CNVm84vEFGrFy832LCHlmTSf5d0Ei#scrollTo=hQhSNDrAVafa&line=12&uniqifier=1):
```
from sdv.single_table import GaussianCopulaSynthesizer
from sklearn.model_selection import train_test_split

real_data, metadata = download_demo('single_table', 'fake_hotel_guests')

# Use train_test_split to have a training data set and a holdout set
train_df, holdout_df = train_test_split(real_data, test_size=0.2)
synthesizer = GaussianCopulaSynthesizer(metadata)
synthesizer.fit(train_df)
synthetic_data = synthesizer.sample(1000)

num_rows_subsample = 100
num_iterations = 1
# SDMetrics uses dictionary of sinlge table metadata and not the Metadata class
metadata_dict = metadata._convert_to_single_table().to_dict()
compute_breakdown_result = DCROverfittingProtection.compute_breakdown(
    train_df, synthetic_data, holdout_df, metadata_dict, num_rows_subsample,num_iterations
)
```

lajohn4747:
resolves #737 
CU-86b48bb5w

`calculate_dcr` already calls `_process_data_with_metadata` so no need to call it in the `_validate_inputs`. This caused the warnings because we already remove the columns when calling `_process_data_with_metadata` the first time.

Add annotation to make warnings fail the test so we don't produce the warnings in the future.

### DIFF
--- a/sdmetrics/single_table/privacy/dcr_baseline_protection.py
+++ b/sdmetrics/single_table/privacy/dcr_baseline_protection.py
@@ -30,7 +30,6 @@ class DCRBaselineProtection(SingleTableMetric):
         cls,
         real_data,
         synthetic_data,
-        metadata,
         num_rows_subsample,
         num_iterations,
     ):
@@ -45,10 +44,13 @@ class DCRBaselineProtection(SingleTableMetric):
             num_rows_subsample = None
             num_iterations = 1
 
-        real_data_copy = real_data.copy()
-        synthetic_data_copy = synthetic_data.copy()
+        if not (isinstance(real_data, pd.DataFrame) and isinstance(synthetic_data, pd.DataFrame)):
+            raise TypeError(
+                f'Both real_data ({type(real_data)}) and synthetic_data ({type(synthetic_data)}) '
+                'must be of type pandas.DataFrame.'
+            )
 
-        return real_data_copy, synthetic_data_copy, num_rows_subsample, num_iterations
+        return num_rows_subsample, num_iterations
 
     @classmethod
     def compute_breakdown(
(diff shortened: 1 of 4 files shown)


sdmetrics/single_table/privacy/dcr_baseline_protection.py
```diff
@@ -48,8 +47,6 @@ def _validate_inputs(
 
         real_data_copy = real_data.copy()
         synthetic_data_copy = synthetic_data.copy()
-        real_data_copy = _process_data_with_metadata(real_data_copy, metadata, True)
-        synthetic_data_copy = _process_data_with_metadata(synthetic_data_copy, metadata, True)
 
         return real_data_copy, synthetic_data_copy, num_rows_subsample, num_iterations
```
frances-h, line 51: Since we're no longer processing the data in this method, can we just return the validated `num_rows_subsample` and `num_iterations`? Also, can we add a check that both the real and synthetic dataframe are pandas dataframes?

frances-h (changes requested):

frances-h (approved):

PR #758, opened 2025-04-01
Produce empty plot if all values are null for `get_column_plot`

### LINKED ISSUE #757: Visualization `get_column_plot` crashes if I have a column of all null values
### Environment Details
* SDMetrics version: 0.19.0
* Python version: 3.11
* Operating System: Linux (Google Colab)

### Error Description
The visualization for `get_column_plot` crashes whenever I have a column of all null values (eg. all `np.nan` or all `None`). In practice, this visualization isn't really that useful. However, I would expect at least that the visualization shouldn't crash. 

I can run into this issue if I am programmatically creating visualizations for all columns of my dataset. If I happen to have a column that is all null, then my script will craash.

### Steps to reproduce
Here I'm trying to plot a column that has only nan values (in both the real and synthetic data). 

```python
import numpy as np
import pandas as pd

from sdmetrics.visualization import get_column_plot

data = pd.DataFrame(data={'A': [np.nan]*100})

get_column_plot(real_data=data,
                synthetic_data=data,
                column_name='A',
                plot_type='bar')
```
```
IndexError: tuple index out of range
```

### Expected Behavior
Rather than crashing, I expect that a visualization can still be rendered. The graph portion can be blank, but the title can still indicate the name of the column and the sub-title can indicate that there are 100% missing values in this column.  Even better, the graph portion can have text that indicates there is no data to graph (if plotly allows this).

A mockup is shown below.

Image

### Additional Context
- The above visualization can be rendered for either of the plot types (`'bar'` or `'distplot'`)
- If either the real or synthetic data is shown to be `None`, then the subtitle and legend would change to exclude the dataset that is not being visualized. But everything else can stay the same.

lajohn4747:
resolves #757 
CU-86b4e3whq

Handle when there is nan values only for both `distplot` and `bar` types.

Updated Font size
Plot

### DIFF
--- a/sdmetrics/visualization.py
+++ b/sdmetrics/visualization.py
@@ -337,6 +337,7 @@ def _generate_column_plot(
             'y': 0.5,
             'showarrow': False,
             'text': 'No data to visualize',
+            'font': {'size': PlotConfig.FONT_SIZE * 2},
         })
 
     show_missing_values = missing_data_real > 0 or missing_data_synthetic > 0
--- a/sdmetrics/visualization.py
+++ b/sdmetrics/visualization.py
@@ -227,11 +227,16 @@ def _generate_column_distplot(real_data, synthetic_data, plot_kwargs={}):
         'colors': colors,
     }
 
-    fig = ff.create_distplot(
-        hist_data,
-        col_names,
-        **{**default_distplot_kwargs, **plot_kwargs},
-    )
+    has_data = any(len(data) > 0 for data in hist_data)
+
+    if has_data:
+        fig = ff.create_distplot(
+            hist_data,
+            col_names,
+            **{**default_distplot_kwargs, **plot_kwargs},
+        )
+    else:
+        fig = go.Figure()
 
     return fig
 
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b4e3whq SDMetrics - Visualization `get_column_plot` crashes if I have a column of all null values #757](https://app.clickup.com/t/86b4e3whq)

sdmetrics/visualization.py
```diff
@@ -227,11 +227,16 @@ def _generate_column_distplot(real_data, synthetic_data, plot_kwargs={}):
...
+    if has_data:
+        fig = ff.create_distplot(
+            hist_data,
+            col_names,
+            **{**default_distplot_kwargs, **plot_kwargs},
+        )
+    else:
+        fig = go.Figure()
 
     return fig
```
pvk-developer, line 241:
```suggestion
    if has_data:
        return ff.create_distplot(
            hist_data,
            col_names,
            **{**default_distplot_kwargs, **plot_kwargs},
        )
        
    return go.Figure()
```

sdmetrics/visualization.py
```diff
@@ -317,13 +322,24 @@ def _generate_column_plot(
...
-        )
+    annotations = []
+    if fig.data:
+        for i, name in enumerate(col_names):
+            fig.update_traces(
+                x=pd.to_datetime(fig.data[i].x) if is_datetime_sdtype else fig.data[i].x,
+                hovertemplate=f'<b>{name}</b><br>Frequency: %{{y}}<extra></extra>',
+                selector={'name': name},
+                **trace_args,
+            )
```
pvk-developer, line 333: can we rename the `i` to `idx` or `index` just to avoid having single characters

npatki:
@lajohn4747 just eyeballing this one -- the font sizes seem a little off. Are they the same sizes that we usually render for plots?

Also, could we increase the font size of the "No data to visualize" sentence too?

lajohn4747:
> @lajohn4747 just eyeballing this one -- the font sizes for all the titles, axes, etc. seem a little off (too small). Are they the same sizes that we usually render for plots?
> 
> Also, could we increase the font size of the "No data to visualize" sentence too?

Yep updated font size.

I don't think anything change, maybe the github rendering might appear weird. Here is a graph from the main branch and the font sizes are the same when I save them.

Original

pvk-developer (approved):

frances-h (approved):

PR #762, opened 2025-04-07
Improve performance of `DCR` metrics

### LINKED ISSUE #760: Improve performance of `DCR` metrics
### Environment Details
* Python version: Python 3.11
* Operating System: Linux Colab

### Problem Description
If I try to apply the DCR metric on a larger dataset (eg. such as one with ~30K rows), the DCR metric currently crashes because it runs out of RAM. This happens _even_ if I apply the `num_rows_subsample` parameter which is meant to reduce the computation. 

The example below illustrates this problem on a demo dataset from SDV that has 32.5K rows (and 19 columns).

```python
from sdmetrics.single_table import DCRBaselineProtection

    modality='single_table',
    dataset_name='census_extended')

metadata_dict = metadata.to_dict()['tables']['census_extended']
metadata_dict['columns']['dob']['datetime_format'] = '%Y-%m-%d'

score = DCRBaselineProtection.compute_breakdown(
  real_data=data,
  synthetic_data=data.copy(), # we don't have synthetic data so I'm passing the same thing in
  metadata=metadata_dict,
  num_rows_subsample=1000)
```

### Expected behavior
The following 3 changes will improve the performance:

1. When computing DCR, both datasets should be chunked. Currently we are not chunking the reference dataset ([See code](https://github.com/sdv-dev/SDMetrics/blob/dcr_feature_branch/sdmetrics/single_table/privacy/dcr_utils.py#L102)).
    - Both datasets should be chunked at 1K rows, meaning that the full cross join will only ever be 1M rows. We will then need to compute DCR over all combinations of chunks.
    - For a given row of data, the DCR value is the _minimum_ distance across all chunks of the reference dataset
2. The `CHUNK_SIZE` should be available as a class-level attribute in the case that a user may want to change this. Eg. The code below would change it to 2500 for future `compute_breakdown` or `compute` calls of the metric.
```python
>>> DCRBaselineProtection.CHUNK_SIZE
1000 # default
>>> DCRBaselineProtection.CHUNK_SIZE = 2500
```
3. The `num_rows_subsample` should subsample all involved datasets, not just the synthetic data.
    - Our initial definition was to subsample only the synthetic data so that we could calculate real DCRs for each of those rows. Now that the other datasets will be subsampled too, the DCRs will all be estimates -- _which is ok for the purposes of speeding up implementation_.
    - Note that if there are multiple iterations, then each iteration should be computed with _different_ subsamples.

frances-h:
Resolve #760

### DIFF
--- a/sdmetrics/single_table/privacy/dcr_baseline_protection.py
+++ b/sdmetrics/single_table/privacy/dcr_baseline_protection.py
@@ -23,6 +23,7 @@ class DCRBaselineProtection(SingleTableMetric):
     goal = Goal.MAXIMIZE
     min_value = 0.0
     max_value = 1.0
+    CHUNK_SIZE = 1000
     _seed = None
 
     @classmethod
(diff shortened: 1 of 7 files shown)


amontanez24 (approved):
LGTM!

pvk-developer (approved):

PR #765, opened 2025-04-14
Fixing sdmetrics import and adding install and import test workflow

### LINKED ISSUE #764: Importing sdmetrics crashes because of torch dependency
### Environment Details

* SDMetrics version: 0.20.0
* Python version:
* Operating System:

### Error Description

Torch is being imported in the file below after [this change](https://github.com/sdv-dev/SDMetrics/pull/736). Since it isn't listed as a required dependency, it crashes whenever you import from sdmetrics unless you happened to already have torch installed.
https://github.com/sdv-dev/SDMetrics/blob/89f2df225564ff636736930567a386785c0f8a18/sdmetrics/single_table/bayesian_network.py#L7

### Steps to reproduce

1. Create a new environment
2. Install sdmetrics and nothing else
3. Run

```
import sdmetrics
```

### Additional Context
- We should wrap the importing of torch in a try/catch like we do here
https://github.com/sdv-dev/SDMetrics/blob/89f2df225564ff636736930567a386785c0f8a18/sdmetrics/timeseries/ml_scorers.py#L28-L33
- We should add an install test workflow like we have [here](https://github.com/sdv-dev/SDV/blob/main/.github/workflows/install.yaml)

amontanez24:
resolves #764

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -1,0 +1,38 @@
+name: Install Tests
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  install:
+    name: ${{ matrix.python_version }} install
+    strategy:
+      fail-fast: true
+      matrix:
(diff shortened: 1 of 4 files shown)


gsheni (approved):

R-Palazzo (approved):
LGTM!

PR #767, opened 2025-04-16
Add a violin plot visualizations to compare a pair of columns

### LINKED ISSUE #759: Add a violin plot visualizations to compare a pair of columns
### Problem Description
I want to visualize a pair of columns where one of the column is categorical and the other is numerical. Currently, the [`get_column_pair_plot` function](https://docs.sdv.dev/sdmetrics/reports/visualization-utilities#compare-a-pair-of-synthetic-columns-and-real-columns-2d) offers the `'box'` plot type that allows me to do this.

For each category value, this box plot displays the quartiles of the numerical column (Q1, median, Q3). While this is useful information, the quartiles may not be enough to fully and intuitively understand the overall shape of the data.

Image

### Expected behavior
To plot the full shape of the data, it would be good to offer a `violin` plot type. A violin plot essentially shows off a full distplot for each category value. Below is a violin plot for the _same data_.

Image

I expect that I should be able to generate such a plot by supplying `plot_type='violin'` into this function.

```python
from sdmetrics.visualization import get_column_pair_plot

fig = get_column_pair_plot(
    real_data=real_table,
    synthetic_data=synthetic_table,
    column_names=['tip', 'smoker'],
    plot_type='violin')

fig.show()
```

Note that all the styling of this visualization (colors, font weights, etc.) should be the same as other plots. This plot should also work if only the real or only the synthetic data is available.

### Additional context
See [Violin plots in Plotly](https://plotly.com/python/violin/) for more info. To render the above visualization, we used `box=False` and `violinmode='overlay'`.

R-Palazzo:
CU-86b4g8r3b
Resolve #759

Here are the visuals; let me know if they work. (For Figs 1 and 3, I placed my mouse over the violin to see what is shown.) @npatki
Screenshot 2025-04-16 at 16 12 07
Screenshot 2025-04-16 at 16 12 42
Screenshot 2025-04-16 at 16 13 23

### DIFF
--- a/sdmetrics/visualization.py
+++ b/sdmetrics/visualization.py
@@ -154,7 +154,7 @@ def _generate_box_plot(all_data, columns):
 
 
 def _generate_violin_plot(data, columns):
-    """Return a violin plot of the real and synthetic data for a given column pair."""
+    """Return a violin plot for a given column pair."""
     fig = px.violin(
         data,
         x=columns[0],
--- a/sdmetrics/visualization.py
+++ b/sdmetrics/visualization.py
@@ -153,6 +153,33 @@ def _generate_box_plot(all_data, columns):
     return fig
 
 
+def _generate_violin_plot(data, columns):
+    """Return a violin plot of the real and synthetic data for a given column pair."""
+    fig = px.violin(
+        data,
+        x=columns[0],
+        y=columns[1],
+        box=False,
+        violinmode='overlay',
+        color='Data',
+        color_discrete_map={
+            'Real': PlotConfig.DATACEBO_DARK,
+            'Synthetic': PlotConfig.DATACEBO_GREEN,
+        },
+    )
+
+    unique_values = data['Data'].unique()
+    title = ' vs. '.join(unique_values)
+    title += f" Data for columns '{columns[0]}' and '{columns[1]}'"
+    fig.update_layout(
+        title=title,
+        plot_bgcolor=PlotConfig.BACKGROUND_COLOR,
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b4g8r3b SDMetrics - Add a violin plot visualizations to compare a pair of columns #759](https://app.clickup.com/t/86b4g8r3b)

npatki (commented):
Visuals are looking good!

As discussed in the eng meeting today, could you check what happens if there are different proportions of real vs. synthetic data? For example, if you sample synthetic data to be 5x the size of the real data, will the violin plot for synthetic data be much fatter than the real data? Ideally we don't want that. We want the plots to be normalized within the respective dataset (real or synthetic).

pvk-developer (approved):
LGTM!

R-Palazzo:
> Visuals are looking good!
> 
> As discussed in the eng meeting today, could you check what happens if there are different proportions of real vs. synthetic data? For example, if you sample synthetic data to be 5x the size of the real data, will the violin plot for synthetic data be much fatter than the real data? Ideally we don't want that. We want the plots to be normalized within the respective dataset (real or synthetic).

Hi @npatki thanks for pointing this out.
Based on the attached screenshot, the data is normalized (here the synthetic data is 100 times the size of the real data).
Screenshot 2025-04-17 at 14 22 52

npatki:
Great @R-Palazzo -- looks good to me!

rwedge (approved):

PR #775, opened 2025-06-17
Add `EqualizedOddsImprovement` metric

### LINKED ISSUE #772: Add a fairness metric that computes Equalized Odds
### Problem Description
While SDMetrics currently includes privacy metrics, it does not have any fairness metrics. We should add an **Equalized Odds** fairness metrics that computes whether there is a bias against a particular group of records.

### Expected behavior
Create a new metric called **EqualizedOddsImprovement** that measures whether the synthetic data improves fairness vs the real data. The outputted score should still be within the [0, 1] range; the value should tell us the direction and degree of improvement:
- score > 0.5 indicates that the synthetic data is more fair than the real data
- score < 0.5 indicates that the synthetic data is less fair than the real data

As an example, consider a loan application dataset where we'd like to predict whether each application will be approved. The Equalized Odds metric will tell us whether the predictions are biased towards a certain group of applications -- such as applications where `requestor_race='Asian'`.

```python
from sdmetrics.single_table import EqualizedOddsImprovement

EqualizedOddsImprovement.compute_breakdown(
  real_training_data=real_dataset,
  synthetic_data=synthetic_dataset,
  real_validation_data=test_dataset,
  metadata=my_metadata,
  prediction_column_name='loan_approved',
  positive_class_label='True',
  sensitive_column_name='requestor_race',
  sensitive_column_value='Asian',
  classifier='XGBoost',
)
```

The breakdown should provide the final score, as well as the sub-scores within the real data vs. the synthetic data. For each, we should provide the prediction counts for both strata (`Asian=True` and `Asian=False`).

```python
{
  'score': 0.7189,
  'real_training_data': {
    'equalized_odds': 0.96033
      'Asian=True': {
        'true_positive': XX,
        'false_positive': XX,
        'true_negative': XX,
        'false_negative': XX
      },
      'Asian=False': {
        'true_positive': XX,
        'false_positive': XX,
        'true_negative': XX,
        'false_negative': XX
      }
    }
  },
  'synthetic_data': {
    # same keys/values as the real training_data
  }
}
```

### Additional context
More details about this metric can be found [in this doc](https://docs.google.com/document/d/1aKhdpFcsf9ktK3njyzL9NW6CGS6aNtJKNSUetWi6vuU/edit?tab=t.0#heading=h.ass1gfyc4z5x) (visible to DataCebo team members only). Once completed, we will publish the methodology and the details within the [SDMetrics documentation](https://docs.sdv.dev/sdmetrics).

fealho:
CU-86b5ayqd6, Resolve #772

### DIFF
--- a/sdmetrics/single_table/equalized_odds.py
+++ b/sdmetrics/single_table/equalized_odds.py
@@ -1,0 +1,462 @@
+# flake8: noqa
+"""EqualizedOddsImprovement metric for single table datasets."""
+
+import pandas as pd
+from sklearn.metrics import confusion_matrix
+
+from sdmetrics.goal import Goal
+from sdmetrics.single_table.base import SingleTableMetric
+from sdmetrics.single_table.utils import (
+    _validate_classifier,
+    _validate_column_consistency,
+    _validate_column_values_exist,
+    _validate_data_and_metadata,
+    _validate_prediction_column_name,
+    _validate_required_columns,
+    _validate_sensitive_column_name,
+    _validate_tables,
+)
+from sdmetrics.single_table.data_augmentation.utils import (
+    _process_data_with_metadata_ml_efficacy_metrics,
+)
+
+from xgboost import XGBClassifier
+
+
+class EqualizedOddsImprovement(SingleTableMetric):
+    """EqualizedOddsImprovement metric.
+
+    This metric evaluates fairness by measuring equalized odds - whether the
+    True Positive Rate (TPR) and False Positive Rate (FPR) are the same
+    across different values of a sensitive attribute.
+
+    The metric compares the equalized odds between real training data and
+    synthetic data, both evaluated on a holdout validation set.
+    """
+
+    name = 'EqualizedOddsImprovement'
+    goal = Goal.MAXIMIZE
+    min_value = 0.0
+    max_value = 1.0
+
+    @classmethod
+    def _validate_data_sufficiency(
+        cls,
+        data,
+        prediction_column_name,
+        sensitive_column_name,
+        positive_class_label,
+        sensitive_column_value,
+    ):
+        """Validate that there is sufficient data for training."""
+        # Create binary versions of the columns
+        prediction_binary = data[prediction_column_name] == positive_class_label
+        sensitive_binary = data[sensitive_column_name] == sensitive_column_value
+
+        # Check both sensitive groups (target value and non-target value)
+        for is_sensitive_group in [True, False]:
+            group_predictions = prediction_binary[sensitive_binary == is_sensitive_group]
+            group_name = 'sensitive' if is_sensitive_group else 'non-sensitive'
+
+            if len(group_predictions) == 0:
+                raise ValueError(f'No data found for {group_name} group.')
+
+            positive_count = group_predictions.sum()
+            negative_count = len(group_predictions) - positive_count
+
+            if positive_count < 5 or negative_count < 5:
+                raise ValueError(
+                    f'Insufficient data for {group_name} group: {positive_count} positive, '
+                    f'{negative_count} negative examples (need ≥5 each).'
+                )
+
+    @classmethod
+    def _preprocess_data(
+        cls,
+        data,
+        prediction_column_name,
+        positive_class_label,
+        sensitive_column_name,
+        sensitive_column_value,
+        metadata,
+    ):
+        """Preprocess the data for binary classification."""
+        data = data.copy()
+
+        # Convert prediction column to binary
+        data[prediction_column_name] = (
+            data[prediction_column_name] == positive_class_label
+        ).astype(int)
+
+        # Convert sensitive column to binary
+        data[sensitive_column_name] = (
+            data[sensitive_column_name] == sensitive_column_value
+        ).astype(int)
+
+        # Handle categorical columns for XGBoost
+        for column, column_meta in metadata['columns'].items():
+            if (
+                column in data.columns
+                and column_meta.get('sdtype') in ['categorical', 'boolean']
+                and column != prediction_column_name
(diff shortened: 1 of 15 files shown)


sdv-team:
Task linked: [CU-86b5ayqd6 SDMetrics - Add a fairness metric that computes Equalized Odds #772](https://app.clickup.com/t/86b5ayqd6)

tests/unit/single_table/test_equalized_odds.py
```diff
@@ -0,0 +1,489 @@
...
+            'synthetic_data': 0.8,
+        }
+        assert abs(result['score'] - expected_result['score']) < 1e-10
+        assert result['real_training_data'] == expected_result['real_training_data']
+        assert result['synthetic_data'] == expected_result['synthetic_data']
+
+    @patch.object(EqualizedOddsImprovement, 'compute_breakdown')
+    def test_compute_returns_score_from_breakdown(self, mock_compute_breakdown):
+        """Test compute method returns just the score from compute_breakdown."""
+        mock_compute_breakdown.return_value = {
```
R-Palazzo, line 478: `# Setup`

tests/unit/single_table/test_equalized_odds.py
```diff
@@ -0,0 +1,489 @@
...
+            'score': 0.75,
+            'real_training_data': 0.6,
+            'synthetic_data': 0.9,
+        }
+
+        result = EqualizedOddsImprovement.compute(
+            pd.DataFrame(), pd.DataFrame(), pd.DataFrame(), {}, 'pred', 'pos', 'sens', 'val'
```
R-Palazzo, line 485: `# Run`

tests/unit/single_table/test_equalized_odds.py
```diff
@@ -0,0 +1,489 @@
...
+        )
+
+        assert result == 0.75
```
R-Palazzo, line 488: `# Assert`

sdmetrics/single_table/equalized_odds.py
```diff
@@ -0,0 +1,465 @@
...
+            if len(group_predictions) == 0:
+                raise ValueError(f'No data found for {group_name} group.')
+
+            positive_count = group_predictions.sum()
+            negative_count = len(group_predictions) - positive_count
+
+            if positive_count < 5 or negative_count < 5:
+                raise ValueError(
+                    f'Insufficient data for {group_name} group: {positive_count} positive, '
+                    f'{negative_count} negative examples (need ≥5 each).'
```
R-Palazzo, line 66: Do we need a for loop here since we are counting both positive and negative?

sdmetrics/single_table/equalized_odds.py
```diff
@@ -0,0 +1,465 @@
...
+
+        # Convert prediction column to binary
+        data[prediction_column_name] = (
+            data[prediction_column_name] == positive_class_label
+        ).astype(int)
+
+        # Convert sensitive column to binary
+        data[sensitive_column_name] = (
+            data[sensitive_column_name] == sensitive_column_value
+        ).astype(int)
```
R-Palazzo, line 90: Does it work if the sensitive_column_value is `np.nan`?

sdmetrics/single_table/equalized_odds.py
```diff
@@ -0,0 +1,465 @@
...
+            sensitive_column_name,
+        )
+
+        synthetic_results = cls._evaluate_dataset(
+            synthetic_processed,
+            real_validation_processed,
+            prediction_column_name,
+            sensitive_column_name,
+        )
+
```
R-Palazzo, line 402: Instead of calling each method twice with the real and synthetic data, could we put this logic in a for loop?

sdmetrics/single_table/equalized_odds.py
```diff
@@ -0,0 +1,465 @@
...
+            1, false_group['true_positive'] + false_group['false_negative']
+        )
+
+        # Compute FPR for each group
+        fpr_true = true_group['false_positive'] / max(
+            1, true_group['false_positive'] + true_group['true_negative']
+        )
+        fpr_false = false_group['false_positive'] / max(
+            1, false_group['false_positive'] + false_group['true_negative']
+        )
```
R-Palazzo, line 179: Since we're doing the same operation multiple times, could we put it in a for loop?

Also, could you let a comment describing how tpr_true is calculated? I couldn't find a reference for the normalization.
```
tpr_true = true_group['true_positive'] / max(
            1, true_group['true_positive'] + true_group['false_negative']
        )
```

R-Palazzo (commented):
This is looking good!

Could you add a integration where the `sensitive_column_value` is `np.nan`

sdmetrics/single_table/equalized_odds.py
fealho: A good question. If the column is categorical the nans are just another category, since the column dtype is object, so the user should pass 'nan', 'None', or whatever the string representation they are using for missing values is.

It won't reach the line of code you are referring if the user passes np.nan, since the data validation before this will complain np.nan is not present in the data.

sdmetrics/single_table/equalized_odds.py
fealho: True positive rate is also called [sensitivity](https://en.wikipedia.org/wiki/Sensitivity_and_specificity).

sdmetrics/single_table/equalized_odds.py
fealho: Yes. There are two groups (1) data that matches the sensitive value and (2) data that doesn't. For both cases we need at least 5 True target values and 5 False target values.

R-Palazzo (approved):
LGTM!

frances-h (approved):
Nice work!

PR #822, opened 2026-01-15
Add a threshold to the CorrelationSimilarity metric

### LINKED ISSUE #816: Add a threshold to the CorrelationSimilarity metric
### Problem Description
In the context of evaluating synthetic data, we care more about whether the synthetic data has preserved trends that were _strongly present in the real data_. If the real data didn't have a strong trend to begin with, then it's fairly typical for synthetic data to also not have a trend; synthesizer's don't usually invent correlations when there is none to begin with. So this case is uninteresting.

Right now, if the real data doesn't have any strong trends, then the [CorrelationSimilarity](https://docs.sdv.dev/sdmetrics/data-metrics/quality/correlationsimilarity#usage) metric will typically report a very high score -- as the synthetic data would also not have such a strong trend. Instead of this, it would be better if the metric itself could be set up with a threshold. If that threshold is not met for the real data (aka there is no strong trend to begin with), then the metric would return a `NaN` instead.

### Expected behavior
In CorrelationSimilarity, add a parameter called `real_correlation_threshold`. The metric would then work as follows:

1. It would compute the correlation on the real data
2. If the _absolute value of the real data's correlation_ exceeds the threshold, then the correlation is considered "strong" and the rest of the metric computation continues.
3. Otherwise, the metric score would be a `NaN` instead. There is no need to compute the correlation on the synthetic data.

The default value of this parameter can be 0 (meaning that the behavior is the same as status quo), however it will be easy for the user (or a report) to set a new value when running it.

```python
from sdmetrics.column_pairs import CorrelationSimilarity

CorrelationSimilarity.compute(
    coefficient='Pearson',
    real_correlation_threshold=0.5
)
```

R-Palazzo:
Resolve #816
CU-86b83kh51

### DIFF
--- a/sdmetrics/column_pairs/statistical/correlation_similarity.py
+++ b/sdmetrics/column_pairs/statistical/correlation_similarity.py
@@ -65,6 +65,12 @@ class CorrelationSimilarity(ColumnPairsMetric):
                 The values from the real dataset.
             synthetic_data (Union[numpy.ndarray, pandas.Series]):
                 The values from the synthetic dataset.
+            coefficient (str):
+                The correlation coefficient to use. Either 'Pearson' or 'Spearman'.
+                Default is 'Pearson'.
+            real_correlation_threshold (float):
+                The minimum absolute correlation value for the real data to be considered
+                correlated. Default is 0.
 
         Returns:
             dict:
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b83kh51 SDMetrics - Add a threshold to the CorrelationSimilarity metric #816](https://app.clickup.com/t/86b83kh51)

pvk-developer (approved):

PR #823, opened 2026-01-15
Add a threshold to the ContigencySimilarity metric 

### LINKED ISSUE #817: Add a threshold to the ContingencySimilarity metric
### Problem Description
In the context of evaluating synthetic data, we care more about whether the synthetic data has preserved trends that were strongly present in the real data. If the real data didn't have a strong trend to begin with, then it's fairly typical for synthetic data to also not have a trend; synthesizer's don't usually invent correlations when there is none to begin with. So this case is uninteresting.

Right now, if the real data doesn't have any strong trends, then the [ContingencySimilarity](https://docs.sdv.dev/sdmetrics/data-metrics/quality/contingencysimilarity) metric will typically report a very high score -- as the synthetic data would also not have such a strong trend. Instead of this, it would be better if the metric itself could be set up with a threshold. If that threshold is not met for the real data (aka there is no strong trend to begin with), then the metric would return a NaN instead.

### Expected behavior
In ContingencySimilarity, add a parameter called `real_association_threshold`. The metric would then work as follows:

1. It would create the contingency table for the real data and compute its [Cramer's association score](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.contingency.association.html)
2. If the association score of the real data exceeds the threshold, then the trend is considered "strong" and the rest of the metric computation continues.
3. Otherwise, the metric score would be NaN instead. There is no need to compute the contingency table on the synthetic data.

The default value of this parameter can be 0 (meaning that the behavior is the same as status quo), however it will be easy for the user (or a report) to set a new value when running it.

```python
from sdmetrics.column_pairs import ContingencySimilarity

ContingencySimilarity.compute(
    num_rows_subsample=1000,
    continuous_column_names=['column_1'],
    num_discrete_bins=10,
    real_association_threshold=0.3
)
```

fealho:
CU-86b83kq9t, Resolve #817

### DIFF
--- a/sdmetrics/column_pairs/statistical/contingency_similarity.py
+++ b/sdmetrics/column_pairs/statistical/contingency_similarity.py
@@ -63,10 +63,9 @@ class ContingencySimilarity(ColumnPairsMetric):
         if (
             not isinstance(real_association_threshold, (int, float))
             or real_association_threshold < 0
+            or real_association_threshold > 1
         ):
-            raise ValueError(
-                '`real_association_threshold` must be a number greater than or equal to zero.'
-            )
+            raise ValueError('real_association_threshold must be a number between 0 and 1.')
 
     @classmethod
     def compute_breakdown(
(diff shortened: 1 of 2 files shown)


sdmetrics/column_pairs/statistical/contingency_similarity.py
```diff
@@ -53,6 +60,14 @@ def _validate_inputs(
             if not isinstance(num_rows_subsample, int) or num_rows_subsample <= 0:
                 raise ValueError('`num_rows_subsample` must be an integer greater than zero.')
 
+        if (
+            not isinstance(real_association_threshold, (int, float))
+            or real_association_threshold < 0
```
R-Palazzo: Should we also validate that it's not above 1?

PR #877, opened 2026-06-25
Add a unified DiagnosticReport and a unified QualityReport

### LINKED ISSUE #874: Consolidate single and multi-table reports into one
### Problem Description
Currently, the Diagnostic and Quality Reports each have a single-table and multi-table version. These versions are identical except for a few minor differences:
- The single-table version accepts the data as a pandas DataFrame whereas the multi-table one accepts it as a dictionary
- The multi-table version typically has some additional properties that are computed (based on the relationships between tables)
- The multi-table version shows the table name in the results (when getting details)

These differences are minor and should be consolidated for simplicity.

### Expected behavior
Consolidate the single vs. multi-table reports into a unified `DiagnosticReport` and a unified `QualityReport`.  The required inputs and outputs should be aligned with the _multi-table_ definition.

In the case of a single table, it really doesn't matter what the table name is, but the table name will be given in the metadata.

I expect that that the consolidated reports should be available within the main `sdmetrics.reports` namespace.

```python
from sdmetric.reports import QualityReport
from sdmetric.reports import DiagnosticReport
```

The old namespace should still work, but raise a `FutureWarning`:
```python
from sdmetric.reports.single_table import QualityReport
```
```
FutureWarning: The single table quality report is deprecated. 
Please use the QualityReport from 'sdmetrics.reports' instead.
```

### Additional context
This issue should be completed after #873, where we will consolidate the metadata.

gsheni:
- Fixes #874
- CU-86bae5vhy

### DIFF
--- a/sdmetrics/reports/multi_table/quality_report.py
+++ b/sdmetrics/reports/multi_table/quality_report.py
@@ -10,6 +10,7 @@ from sdmetrics.reports.multi_table.base_multi_table_report import BaseMultiTable
 from sdmetrics.reports.utils import (
     DEFAULT_REAL_ASSOCIATION_THRESHOLD,
     DEFAULT_REAL_CORRELATION_THRESHOLD,
+    _warn_deprecated_report,
 )
 
 
@@ -21,6 +22,7 @@ class QualityReport(BaseMultiTableReport):
     """
 
     def __init__(self):
+        _warn_deprecated_report('multi table quality', 'QualityReport')
         super().__init__()
         self.real_correlation_threshold = DEFAULT_REAL_CORRELATION_THRESHOLD
         self.real_association_threshold = DEFAULT_REAL_ASSOCIATION_THRESHOLD
--- a/sdmetrics/reports/multi_table/quality_report.py
+++ b/sdmetrics/reports/multi_table/quality_report.py
@@ -7,6 +7,10 @@ from sdmetrics.reports.multi_table._properties import (
     InterTableTrends,
 )
 from sdmetrics.reports.multi_table.base_multi_table_report import BaseMultiTableReport
+from sdmetrics.reports.utils import (
+    DEFAULT_REAL_ASSOCIATION_THRESHOLD,
+    DEFAULT_REAL_CORRELATION_THRESHOLD,
+)
 
 
 class QualityReport(BaseMultiTableReport):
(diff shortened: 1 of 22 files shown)


sdv-team:
Task linked: [CU-86bae5vhy SDMetrics - Consolidate single and multi-table reports into one #874](https://app.clickup.com/t/86bae5vhy)

sdmetrics/reports/multi_table/quality_report.py
```diff
@@ -17,9 +22,10 @@ class QualityReport(BaseMultiTableReport):
     """
 
     def __init__(self):
+        _warn_deprecated_report('multi table quality', 'QualityReport')
```
pvk-developer: My only concern is the warning will be produced when you initialize the QR, I believe that we want something like this: https://github.com/sdv-dev/RDT/blob/main/rdt/transformers/text.py
When you import it then you raise the warning.

gsheni: Gotcha, I moved the FutureWarning so it happens during import:
https://github.com/sdv-dev/SDMetrics/pull/877/commits/8954b716783bfc742a492bfd95d99f6028eb00dc

tests/integration/reports/test_unified_reports.py
```diff
@@ -0,0 +1,528 @@
...
+    real_data, synthetic_data, metadata = single_table_demo
+
+    # Run
+    report = DiagnosticReport()
+    report.generate({'table': real_data}, {'table': synthetic_data}, metadata, verbose=False)
+
+    # Assert
+    expected_properties = pd.DataFrame({
+        'Property': ['Data Validity', 'Data Structure', 'Relationship Validity'],
+        'Score': [1.0, 1.0, float('nan')],
```
amontanez24, line 154: should the non multi table properties show as nan or not at all?

amontanez24 (commented):
Looks good! One small comment

tests/integration/reports/test_unified_reports.py
gsheni: The non multi table properties should show up as NaN. I verified this in [slack](https://datacebo.slack.com/archives/C01QNE9MLUB/p1782483597736589?thread_ts=1782482086.338829&cid=C01QNE9MLUB) with Neha.


> **Gaurav**: If the user passes single-table data to the unified QualityReport, the score will be NaN for Cardinality and Intertable trends (seen in the verbose output). Is that the intended behavior?

> **Neha**: Yup that’s right!

amontanez24 (approved):

pvk-developer (approved):

PR #878, opened 2026-06-26
Consolidate single-table and multi-table metadata into one

### LINKED ISSUE #873: Consolidate single and multi-table metadata into one
### Problem Description
Currently, SDMetrics still makes a distinction between single- and multi-table metadata. The multi-table metadata has a dictionary that maps out each table name to the table's metadata; the single-table metadata skips this and directly has the columns.

In the rest of the SDV ecosystem, we have consolidated both of these into a unified metadata object.

### Expected behavior
SDMetrics should also consolidate the single- and multi-table metadata into a unified metadata dictionary definition. Similar to SDV, the metadata in SDMetrics should adhere to the _multi-table_ format. When applied to single tables, it doesn't really matter what the the table name is.

### Metrics/Reports
We should also make sure that all metrics/reports requiring metadata are updated.
- Metrics/reports requiring multi-table metadata should continue to work
- Metrics/reports requiring single-table metadata would need some adjustments as there is now one more level of nesting in the metadata dict.

pvk-developer:
Resolves #873 
CU-86bae5rfx


sdmetrics/demos/single_table/metadata.json
```diff
@@ -1,68 +1,72 @@
...
-        "mba_perc": {
-            "sdtype": "numerical",
-            "computer_representation": "Float"
-        },
-        "work_experience": {
-            "sdtype": "boolean"
-        },
-        "degree_type": {
-            "sdtype": "categorical"
+    "tables": {
```
gsheni, line 2: Do we want to keep this old metadata.json (for integration testing)? That way, we can make sure we are being backwards compatible.

sdmetrics/demos.py
```diff
@@ -35,16 +35,11 @@ def load_demo(modality='multi_table'):
...
-            synthetic_data[table] = _load_table(table_meta, demo_path / f'{table}_synthetic.csv')
-
-    else:
-        real_data = _load_table(metadata, demo_path / 'real.csv')
-        synthetic_data = _load_table(metadata, demo_path / 'synthetic.csv')
+    real_data = {}
+    synthetic_data = {}
+    for table, table_meta in metadata['tables'].items():
+        real_data[table] = _load_table(table_meta, demo_path / f'{table}_real.csv')
+        synthetic_data[table] = _load_table(table_meta, demo_path / f'{table}_synthetic.csv')
```
gsheni, line 43: Why do we need to change this? It means that modality is unused in the function

sdv-team:
Task linked: [CU-86bae5rfx SDMetrics - Consolidate single and multi-table metadata into one #873](https://app.clickup.com/t/86bae5rfx)

sdmetrics/demos.py
pvk-developer: modality is still used on line 34:
```python
    demo_path = pathlib.Path(__file__).parent / 'demos' / modality
```

sdmetrics/demos/single_table/metadata.json
pvk-developer: Not really needed to maintain two metadata.json's. We can include testing without this metadata.json (passing in a dictionary with the expected structure).

sdmetrics/_utils_metadata.py
```diff
@@ -17,44 +17,30 @@ def _validate_metadata_dict(metadata):
...
-    """Validate the metadata for multiple tables."""
+def _validate_metadata(metadata):
+    """Validate the metadata."""
     _validate_metadata_dict(metadata)
     if 'tables' not in metadata:
-        raise ValueError(
-            "Multi-table metadata must include a 'tables' key that maps table names"
-            ' to their respective metadata.'
-        )
+        _validate_columns_exist_in_single_table_metadata(metadata)
```
amontanez24, line 34: should we raise a warning saying that this type of metadata is deprecated?

pvk-developer: I would like to avoid that because internally some metrics use the metadata per each table (and we iterate over it as `metadata['tables'].items()`). Within the unified reports we raise an error if the metadata is not the unified one.

gsheni (approved):

frances-h (approved):

PR #883, opened 2026-07-09
Make reports not return skipped properties and improve user facing errors

### LINKED ISSUE #881: Make the unified report errors more user-friendly for single-table datasets
### Problem Description
In https://github.com/sdv-dev/SDMetrics/issues/874, we are consolidating the single- and multi-table reports into a unified report for all modalities. After generating the reports, the user is still able to call `get_visualization` and `get_details` for properties that are skipped for single table data.

The problem is that the output of this case isn't consistent. Eg. Sometimes the details are an empty table and sometimes it's just overall empty; sometimes the visualization crashes with an error that seems to come internally from the graphing library (eg. `UFuncTypeError: : ufunc ‘add’ did not contain a loop…`).

Additionally, `get_properties` currently includes properties that are skipped for single-table data, even though those properties are not evaluated.

### Expected behavior
If I have single-table data and I call `get_details` or `get_visualization` on a property that is not computed for single-table data, then I expect to see a user-friendly error about this.

```
Error: This property is not available for single-table datasets.
```
If I call `get_properties`, I expect it to return only the properties that were actually evaluated. The method is documented as returning properties that the report evaluated, and skipped properties are not evaluated. Since users can’t act on those rows, omitting them is clearer (for single-table data). 

This applies to:
- Diagnostic Report: Relationship Validity
- Quality Report: Cardinality, Intertable Trends

pvk-developer:
Resolve #881

### DIFF
--- a/sdmetrics/reports/base_unified_report.py
+++ b/sdmetrics/reports/base_unified_report.py
@@ -84,3 +84,25 @@ class BaseUnifiedReport(BaseMultiTableReport):
             synthetic_data,
             metadata,
         )
+
+    def get_properties(self):
+        """Return the property score.
+
+        Returns:
+            pandas.DataFrame
+                The property score.
+        """
+        self._check_report_generated()
+        name, score = [], []
+        is_single_table = self.report_info['num_tables'] == 1
+
+        for property_name, property_instance in self._properties.items():
+            if not (is_single_table and property_name in self._SINGLE_TABLE_SKIPPED_PROPERTIES):
+                property_score = property_instance._compute_average()
+                name.append(property_name)
+                score.append(property_score)
+
+        return pd.DataFrame({
+            'Property': name,
+            'Score': score,
+        })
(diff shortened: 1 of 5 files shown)


sdmetrics/reports/base_unified_report.py
```diff
@@ -84,3 +84,25 @@ def _validate(self, real_data, synthetic_data, metadata):
...
+    def get_properties(self):
+        """Return the property score.
+
+        Returns:
+            pandas.DataFrame
+                The property score.
+        """
+        self._check_report_generated()
+        name, score = [], []
+        is_single_table = self.report_info['num_tables'] == 1
```
gsheni: Is this not needed? Or will `report_info` be defined by this point? 
```suggestion
        is_single_table = self.report_info.get('num_tables', 0) == 1
```
It seems as if this is only defined in this situation:
https://github.com/sdv-dev/SDMetrics/blob/42223b16018d582a43dde82239357b85e708f450/sdmetrics/reports/base_report.py#L159

gsheni (approved):

gsheni (approved):

frances-h (approved):

PR #903, opened 2026-08-11
`DatetimeFormatAdherence` metric

### LINKED ISSUE #898: Add `DatetimeFormatAdherence` metric
### Problem Description
This metric captures whether the synthetic data adheres to a specified datetime format. It is meant to be used for datetime columns that have a specific format.

### Expected behavior
- Add a new `single_column` metric that calculates the percent of values that follows the specified format.
- This metric takes in the expected `datetime_format` and is applied to only datetime columns.

#### Attributes
- `name`: `'DatetimeFormatAdherence'`
- `goal`: `Goal.MAXIMIZE`
- `min_value`: 0.0
- `max_value`: 1.0

#### Methods
- `compute(real_data, synthetic_data, datetime_format)`: Compute the score for the metric. The returned score is the percentage of synthetic data values that match the datetime format.  
    - Parameters:
        - (required) `real_data`: A `pandas.Series` object with the column of real data
        - (required) `synthetic_data`: A `pandas.Series` object with the column of synthetic data
        - (required) `datetime_format`: A string containing the datetime format to check against. For more information, see [Python's strftime module](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes).

```python
>>> from sdmetrics.single_column import DatetimeFormatAdherence
>>> DatetimeFormatAdherence.compute(
    real_data=real_table['purchase_date'],
    synthetic_data=synthetic_table['purchase_date'],
    datetime_format='%Y-%m-%d'
)
```

#### Notes
* The datetime format can only be checked if the synthetic data has a string representation. If it is present as datetime64, then the format cannot be checked and the score is NaN.
* If the real data is represented as a string and it does not match the provided datetime format, then the metric will issue a warning but it will continue to check the synthetic data.

sarahmish:
Resolve #898

### DIFF
--- a/sdmetrics/single_column/statistical/datetime_format_adherence.py
+++ b/sdmetrics/single_column/statistical/datetime_format_adherence.py
@@ -101,7 +101,18 @@ class DatetimeFormatAdherence(SingleColumnMetric):
         if is_string_dtype(real_data[~real_data_nan]):
             real_valid_rows = cls._filter_valid_datetime_rows(real_data, datetime_format)
             if len(real_valid_rows) != len(real_data):
-                warnings.warn('The real data does not match the given datetime format.')
+                invalid = real_data[~real_data.index.isin(real_valid_rows.index)]
+                num_examples = 2
+                message = (
+                    'Some values in the real data do not match the specified datetime format: '
+                    + ', '.join(f"'{value}'" for value in invalid.head(num_examples).astype(str))
+                )
+
+                remaining = len(invalid) - num_examples
+                if remaining > 0:
+                    message += f' + {remaining} more.'
+
+                warnings.warn(message)
 
         synthetic_valid_rows = cls._filter_valid_datetime_rows(synthetic_data, datetime_format)
         score = len(synthetic_valid_rows) / len(synthetic_data)
--- a/sdmetrics/single_column/statistical/datetime_format_adherence.py
+++ b/sdmetrics/single_column/statistical/datetime_format_adherence.py
@@ -1,7 +1,7 @@
 """Datetime Format Adherence Metric."""
 
-import logging
 import re
+import warnings
 from datetime import datetime
 
 import numpy as np
@@ -10,8 +10,6 @@ from pandas.api.types import is_string_dtype
 
 from sdmetrics.goal import Goal
 from sdmetrics.single_column.base import SingleColumnMetric
-
-LOGGER = logging.getLogger(__name__)
 
 
 class DatetimeFormatAdherence(SingleColumnMetric):
@@ -52,7 +50,7 @@ class DatetimeFormatAdherence(SingleColumnMetric):
             raise ValueError(message)
 
     @staticmethod
-    def _validate_datetime_column(column, datetime_format):
+    def _filter_valid_datetime_rows(column, datetime_format):
         """Return values from the column that match the specified datetime format.
 
         Args:
@@ -101,11 +99,11 @@ class DatetimeFormatAdherence(SingleColumnMetric):
 
         cls._validate_datetime_format(datetime_format)
         if is_string_dtype(real_data[~real_data_nan]):
-            real_valid_rows = cls._validate_datetime_column(real_data, datetime_format)
+            real_valid_rows = cls._filter_valid_datetime_rows(real_data, datetime_format)
             if len(real_valid_rows) != len(real_data):
-                LOGGER.warning('The real data does not match the given datetime format.')
+                warnings.warn('The real data does not match the given datetime format.')
 
-        synthetic_valid_rows = cls._validate_datetime_column(synthetic_data, datetime_format)
+        synthetic_valid_rows = cls._filter_valid_datetime_rows(synthetic_data, datetime_format)
         score = len(synthetic_valid_rows) / len(synthetic_data)
 
         return score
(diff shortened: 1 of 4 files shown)


tests/unit/single_column/statistical/test_datetime_format_adherence.py
```diff
@@ -0,0 +1,182 @@
...
+        datetime_format = '%d-%m-%Y'
+        expected = pd.Series([True, True, True])
+
+        metric = DatetimeFormatAdherence()
+
+        # Run
+        result = metric._validate_datetime_column(data, datetime_format)
+
+        # Assert
+        result.equals(expected)
```
pvk-developer: missing assert

tests/unit/single_column/statistical/test_datetime_format_adherence.py
```diff
@@ -0,0 +1,182 @@
...
```
pvk-developer: missing assert

sdmetrics/single_column/statistical/datetime_format_adherence.py
```diff
@@ -0,0 +1,111 @@
...
+            formated_date = datetime.now().strftime(datetime_format)
+        except Exception as exception:
+            raise ValueError(message) from exception
+
+        matches = re.findall('(%.)|(%)', formated_date)
+        if matches:
+            raise ValueError(message)
+
+    @staticmethod
+    def _validate_datetime_column(column, datetime_format):
```
pvk-developer: Could we rename this method to  `_filter_valid_datetime_rows` or similar to represent better its functionality ?

sdmetrics/single_column/statistical/datetime_format_adherence.py
```diff
@@ -0,0 +1,111 @@
...
+        real_data_nan = pd.isna(real_data)
+        synthetic_data_nan = pd.isna(synthetic_data)
+        if not is_string_dtype(synthetic_data[~synthetic_data_nan]):
+            return np.nan
+
+        cls._validate_datetime_format(datetime_format)
+        if is_string_dtype(real_data[~real_data_nan]):
+            real_valid_rows = cls._validate_datetime_column(real_data, datetime_format)
+            if len(real_valid_rows) != len(real_data):
+                LOGGER.warning('The real data does not match the given datetime format.')
```
pvk-developer: Should this be a `warnings.warn` ?

tests/unit/single_column/statistical/test_datetime_format_adherence.py
```diff
@@ -0,0 +1,182 @@
...
+        synthetic_data = pd.Series(['9-10-2020', np.nan, '15-12-2020'])
+        datetime_format = '%d-%m-%Y'
+
+        metric = DatetimeFormatAdherence()
+
+        # Run
+        result = metric.compute(real_data, synthetic_data, datetime_format)
+
+        # Assert
+        assert result == 1.0
```
pvk-developer, line 172: Can we add one more test like this that computes a score of `.75`, for example:
```python
synthetic_data = pd.Series([
    "2026-01-01",
    "2026-01-02",
    np.nan,
    "2026-01-02 12:20:59",
])
```

sdmetrics/single_column/statistical/datetime_format_adherence.py
```diff
@@ -0,0 +1,111 @@
...
+        if not isinstance(datetime_format, str):
+            raise ValueError('`datetime_format` must be a string.')
+
+        message = f"Invalid datetime format string '{datetime_format}'."
+        try:
+            formated_date = datetime.now().strftime(datetime_format)
+        except Exception as exception:
+            raise ValueError(message) from exception
+
+        matches = re.findall('(%.)|(%)', formated_date)
```
amontanez24, line 48: Is this checking that the datetime format was applied correctly? How do you get here without strftime erroring?

sarahmish: I pulled it from SDV directly for [datetime validation](https://github.com/sdv-dev/SDV/blob/461537559cd54ec769226738dfba6fbb114a3709/sdv/metadata/single_table.py#L161). The second check is to make sure that no stray `%` are kept in the formated date
```python3
>>> formated_date = datetime.now().strftime("%%")
>>> formated_date # this would be "%"
```
I'm not sure why `strftime` doesn't consider this an improper format though.

amontanez24 (approved):

tests/unit/single_column/statistical/test_datetime_format_adherence.py
sarahmish: This example fails on minimum tests, with pandas 1.5 it considers "2026-01-02 12:20:59" a valid format and generalizes the rest to be the same
```
0   2026-01-01 00:00:00
1   2026-01-02 00:00:00
2                   NaT
3   2026-01-02 12:20:59
```

In pandas > 2, we get what we expect:
```
0   2026-01-01
1   2026-01-02
2          NaT
3          NaT
```

Since the logic is similar to SDV's [datetime column validation](https://github.com/sdv-dev/SDV/blob/461537559cd54ec769226738dfba6fbb114a3709/sdv/_utils.py#L128), it will also consider the first to be valid if you're running the same version of pandas.

I will update the test to be an incomplete score but with another example.

Let me know if you agree with this.

pvk-developer (approved):

PR #904, opened 2026-08-12
Add regex format adherence

### LINKED ISSUE #897: Add `RegexFormatAdherence` metric
### Problem Description
This metric captures whether the synthetic data adheres to a specified Regex format. It is meant to be used for ID columns that have a specific format.

### Expected behavior
- Add a new `single_column` metric that calculates the percent of values that follows the specified format.
- This metric takes in the expected `regex_format` and is applied to only ID columns.

#### Attributes
- `name`: `'RegexFormatAdherence'`
- `goal`: `Goal.MAXIMIZE`
- `min_value`: 0.0
- `max_value`: 1.0

#### Methods
- `compute(real_data, synthetic_data, regex_format)`: Compute the score for the metric. The returned score is the percentage of synthetic data values that match the regex format. (e.g. a score of 0.4 means 40% of the rows match the regex format and 60% do not have the specified format).  
    - Parameters:
        - (required) `real_data`: A `pandas.Series` object with the column of real data
        - (required) `synthetic_data`: A `pandas.Series` object with the column of synthetic data
        - (required) `regex_format`: A string containing the Regex format to check against

```python
>>> from sdmetrics.single_column import RegexFormatAdherence
>>> RegexFormatAdherence.compute(
    real_data=real_table['product_id'],
    synthetic_data=synthetic_table['product_id'],
    regex_format='(?P<country>[A-Z]{2})_[0-9A-Z]{5}'
)
```

#### Notes
* If the Regex format has a named capture group, then we expect the contents of the named group should be found somewhere within the real data in order to count as a match.
  * For example, the Regex format might be: (?P<country>[A-Z]{2})_[0-9A-Z]{5}  which indicates that the first 2 letters are a named capture group representing a country, followed by 5 random alphanumeric digits. For example US_0LKE3 or CA_01W3E.
  * In this case we would expect that the country codes in the synthetic data (US or CA) are exact codes that are present in the real data. (Real data should also contain US or CA.) If the country code cannot be found in the real data, then the format isn't met for that value.

* If the real data doesn't match the provided Regex, then the metric will provide a warning but it will still continue to check the synthetic data. If the real data doesn't match the provided Regex and there is a named capture group, then the metric cannot be computed and the score should be NaN.

sarahmish:
Resolve #897

### DIFF
--- a/sdmetrics/single_column/statistical/regex_format_adherence.py
+++ b/sdmetrics/single_column/statistical/regex_format_adherence.py
@@ -63,7 +63,7 @@ class RegexFormatAdherence(SingleColumnMetric):
         groups = None
         if re.compile(regex_format).groups:
             groups = column[regex_column].str.extract(regex_format)
-            if compare is not None:
+            if compare is not None and not compare.empty:
                 valid_groups = set(map(tuple, compare.dropna().to_numpy()))
                 regex_column = groups.apply(tuple, axis=1).isin(valid_groups)
 
@@ -94,13 +94,13 @@ class RegexFormatAdherence(SingleColumnMetric):
             invalid_values = real_data[~real_data.index.isin(real_valid.index)]
             num_examples = 2
             message = (
-                "Some values in the real data do not match the specified regex format: "
-                + ", ".join(f"'{value}'" for value in invalid_values.head(num_examples).astype(str))
+                'Some values in the real data do not match the specified regex format: '
+                + ', '.join(f"'{value}'" for value in invalid_values.head(num_examples).astype(str))
             )
 
             remaining = len(invalid_values) - num_examples
             if remaining > 0:
-                message += f" + {remaining} more."
+                message += f' + {remaining} more.'
 
             warnings.warn(message)
 
--- a/sdmetrics/single_column/statistical/regex_format_adherence.py
+++ b/sdmetrics/single_column/statistical/regex_format_adherence.py
@@ -52,6 +52,8 @@ class RegexFormatAdherence(SingleColumnMetric):
                 Column to evaluate.
             regex_format (str):
                 The regex format.
+            compare (pd.Series | pd.DataFrame, optional):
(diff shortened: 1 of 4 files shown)


sdmetrics/single_column/statistical/regex_format_adherence.py
```diff
@@ -0,0 +1,101 @@
...
+from sdmetrics.goal import Goal
+from sdmetrics.single_column.base import SingleColumnMetric
+
+LOGGER = logging.getLogger(__name__)
+
+
+class RegexFormatAdherence(SingleColumnMetric):
+    """Regex format adherence  metric.
+
+    The proportion of data points in the synthetic data that match
```
R-Palazzo, line 15:
```suggestion
    The proportion of data points in the synthetic data that matches
```

sdmetrics/single_column/statistical/regex_format_adherence.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Returns:
+            float:
+                The proportion of data points in the synthetic data that match the regex format.
+        """
+        cls._validate_regex_format(regex_format)
+
+        real_valid, real_groups = cls._validate_regex_column(real_data, regex_format)
+        if len(real_valid) != len(real_data):
+            LOGGER.warning('The real data does not match the given regex format.')
```
R-Palazzo: Maybe we could add examples here:
```suggestion
            LOGGER.warning(
    "Some values in the real data do not match the specified regex format: "
    "'abcd', 'cdef' + 5 more."
)
```

R-Palazzo (approved):
LGTM!

sdmetrics/single_column/statistical/regex_format_adherence.py
sarahmish: I think "match" is correct because it refers to "data points"

frances-h (approved):

PR #913, opened 2026-08-20
Constraint adherence metric

### LINKED ISSUE #899: Add `ConstraintAdherence` metric
### Problem Description
This metric captures whether the synthetic data adheres to a constraint.

### Expected behavior
- Add a new `multi_table` metric that checks each data point in the synthetic data to see if it follows the constraint that is provided.
- This metric takes in the constraint in its JSON (dictionary) representation.

#### Attributes
- `name`: `'ConstraintAdherence'`
- `goal`: `Goal.MAXIMIZE`
- `min_value`: 0.0
- `max_value`: 1.0

#### Methods
- `compute(real_data, synthetic_data, metadata, constraint)`: Compute the score for the metric. The returned score is the percentage of synthetic data values that match the constraint definition.  
    - Parameters:
        * (required) `real_data`: A dictionary that maps each table name to a `pandas.DataFrame` object containing the real data for that table
        * (required) `synthetic_data`: A dictionary that maps each table name to a `pandas.DataFrame` object containing the synthetic data for that table
        * (required) `metadata`: A dictionary containing the full metadata describing the data
        * (required) `constraint`: A dictionary that defines the type of constraint and the parameters. This dictionary corresponds to a single constraint only and should follow the [JSON definition from SDV](https://docs.sdv.dev/sdv/modeling/constraint-augmented-generation-cag/constraints-api?select=numerical,datetime,categorical,personal-info,banking,networking,location#json-representation). In short: The `class_name` parameter is the full name of the constraint class, and the parameters dictionary maps each of the constraint parameters (key) to the value.

```python
>>> from sdmetrics.multi_table import ConstraintAdherence
>>> ConstraintAdherence.compute(
    real_data=real_data_dict,
    synthetic_data=synthetic_data_dict,
    metadata=metadata_dict,
    constraint={
      'class_name': 'DenormalizedTable',
      'parameters': {
        'table_name': 'UserTransaction',
        'denormalized_primary_key': 'User ID',
        'denormalized_column_names': ['DateOfBirth', 'FirstName', 'LastName']
      } 
    }
)
```

#### Notes
* If the constraint is completely invalid (e.g. an unknown constraint), then it should be an error. 
* If the constraint cannot be checked for some reason (for example, the referenced table or column names don't exist in the data), then the metric will issue a warning to the debug log and the score will be NaN.
* If the real data does not match the constraint, then the metric will log a warning but it will continue to check the synthetic data.
* The exact logic is based on the constraint. In other words, there will be custom logic based on the `class_name`.
  * Note that some constraints may not really have a logic. For example a DenormalizedTable with only the primary key (and no column names) doesn't have anything to check; in this case, the metric can just return a score of 1.0.

sarahmish:
Resolve #899 

This PR focuses on single-table constraints. I will follow-up with multi-table constraints in another PR.


frances-h (approved):

sdmetrics/multi_table/statistical/constraints/base.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        raise NotImplementedError()
+
+    def is_valid(self, data, metadata=None):
+        """Say whether the given table rows are valid.
+
+        Args:
+            data (pd.DataFrame or dict[pd.DataFrame]):
+                Table data.
+
```
R-Palazzo, line 172: Add `metadata`

sdmetrics/multi_table/statistical/constraints/base.py
```diff
@@ -0,0 +1,219 @@
...
+        Returns:
+            pd.Series or dict[pd.Series]:
+                Series of boolean values indicating if the row is valid for the constraint or not.
+        """
+        metadata = self.metadata if metadata is None else metadata
+        self._validate_data(data, metadata)
+
+        is_valid_data = self._is_valid(data, metadata)
+        if isinstance(data, pd.DataFrame) or self._single_table:
```
R-Palazzo: Checking on this: should the data always be a dict of pd.DataFrame?

sdmetrics/multi_table/statistical/constraints/base.py
```diff
@@ -0,0 +1,219 @@
...
+        constraint_class = constraint_classes[class_name]
+        try:
+            return constraint_class(**constraint_dict['parameters'])
+        except TypeError as ex:
+            raise ValueError(f"Unable to create the constraint '{class_name}': {ex}") from ex
+
+    def __init__(self):
+        self.metadata = None
+        self._fitted = False
+        self._single_table = False
```
R-Palazzo: Do we plan to use this flag?

I didn't see it in the constraints of this PR, and since they are single-table, I was wondering if it's necessary to keep it.

sdmetrics/multi_table/statistical/constraints/_utils.py
R-Palazzo, line 1: Could we add unit tests for those methods?
Maybe we can reuse the existing ones when those methods come from another library.

tests/unit/multi_table/statistical/constraints/test_chained_inequality.py
```diff
@@ -0,0 +1,309 @@
...
+        result_equal = constraint_equal._is_valid(data_float_lower_equal, metadata_float)
+
+        # Assert
+        expected_result = pd.Series([True] * 5)
+        pd.testing.assert_series_equal(result_strict['table'], expected_result)
+        pd.testing.assert_series_equal(result_equal['table'], expected_result)
+
+    def test__is_valid_with_and_without_strict(self):
+        """Test the ``is_valid`` method with and without strict boundaries."""
+        table_data_valid = {
```
R-Palazzo, line 149: `# Setup`

tests/unit/multi_table/statistical/constraints/test_referential_hierarchy.py
```diff
@@ -0,0 +1,494 @@
...
+
+    def test___init__invalid_scaling_method(self):
+        """Test the ``__init__`` method errors with an unknown scaling method."""
+        # Setup
+        err_msg = re.escape(
+            "Unrecognized scaling_method 'wide'. The scaling method must be one of "
+            "'branch', 'depth' or 'multiply'."
+        )
+
+        # Run and Assert
```
R-Palazzo, line 202: Could we keep a single wording for this, either `# Run and Assert` or `# Run & Assert`?

R-Palazzo (commented):
Looking really good!

I left just a few minor comments, mostly around the single- vs multi-table handling.

Could we also add an integration test similar to the code snippet in the issue, so we can check the import and the end-to-end usage as well?

sdmetrics/multi_table/statistical/constraints/base.py
sarahmish: correct, according to the issue it should be. I'll update it.

R-Palazzo (approved):
LGTM!

PR #923, opened 2026-08-31
SDV 2.0: Reports should understand and parse the `ordinal` sdtype

### LINKED ISSUE #908: SDV 2.0: Reports should understand and parse the `ordinal` sdtype
### Problem Description
In SDV 2.0, we are introducing the concept of an `ordinal` sdtype, which is meant for discrete, ordinal data. (Compared to the existing `categorical` sdtype which is meant for unordered, nominal data.)

SDMetrics should be made compatible with this concept.

### Expected behavior
The diagnostic and quality reports should be able to accept metadata that contains the `ordinal` sdtype.

In this initial iteration, the reports should just apply the same metrics as `categorical` sdtypes for `ordinal`. In the future we plan to develop new metrics that are specific to discrete, ordinal data.

R-Palazzo:
Resolve #908
86bbgkbrg

### DIFF
--- a/sdmetrics/single_table/new_row_synthesis.py
+++ b/sdmetrics/single_table/new_row_synthesis.py
@@ -101,7 +101,7 @@ class NewRowSynthesis(SingleTableMetric):
             numerical_fields = []
 
         try:
-            categorical_fields = cls._select_fields(metadata, ('categorical', 'boolean'))
+            categorical_fields = cls._select_fields(metadata, ('categorical', 'boolean', 'ordinal'))
         except IncomputableMetricError:
             categorical_fields = []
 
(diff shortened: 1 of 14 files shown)


sdmetrics/single_table/new_row_synthesis.py
```diff
@@ -101,7 +101,7 @@ def compute_breakdown(
             numerical_fields = []
 
         try:
-            categorical_fields = cls._select_fields(metadata, ('categorical', 'boolean'))
+            categorical_fields = cls._select_fields(metadata, ('categorical', 'boolean', 'ordinal'))
```
R-Palazzo, line 104: I made this change to keep the logic consistent, but since this metric isn't used in the reports, it could be considered out of scope for this issue. Let me know if we can keep the change or if we'd prefer to modify only what's related to the reports.

amontanez24 (approved):

frances-h (approved):

PR #931, opened 2026-09-03
Update datetime columns to perserve their original format

### LINKED ISSUE #925: The DiagnosticReport reports a score of 1.0 even if a datetime format cannot be checked
### Environment Details
* SDMetrics version: 0.30.0
* Python version: 3.12
* Operating System: Linux (Google Colab)

### Error Description
As part of the DataValidity property, the [DatetimeFormatAdherence](https://docs.sdv.dev/sdmetrics/data-metrics/diagnostic/datetimeformatadherence) metric is applied whenever the metadata has a `datetime_format` present. 

However if the data itself is format-less (eg. is something else, like a `datetime64` or `<M8[ns]`), then there is nothing to check. In this case, I would expect the score for this column to be `NaN`; instead I'm seeing a 1.0.

### Steps to reproduce
The datetime columns in the demo right now (eg. `start_date`) are format-less. We will fix this in #926, but at least for now, we should expect the score to be `NaN`.

```python
from sdmetrics.reports import DiagnosticReport

real_data, synthetic_data, metadata = load_demo(modality='single_table')

diagnostic = DiagnosticReport()
diagnostic.generate(real_data, synthetic_data, metadata)
diagnostic.get_details('Data Validity')
```

Instead, obverse that is is 1.0:

Image

### Root Cause
Note that `DatetimeFormatAdherence` metric itself correctly returns a `NaN` for this data.

```python
DatetimeFormatAdherence.compute(
    real_data=real_data['student_placements']['start_date'],
    synthetic_data=synthetic_data['student_placements']['start_date'],
    datetime_format='%Y-%m-%d'
)
```
```
nan
```

The root cause comes from the report itself:
- The report processes all datetime columns and converts them to datetime64. See the conversion code [here](https://github.com/sdv-dev/SDMetrics/blob/4ae714286b47412e032b267277c861fe8cab11d5/sdmetrics/reports/base_report.py#L93).
- In order to check the DatetimeFormatAdherence metric, we later convert these columns back into a string representation to give to the metric alongside the datetime_format value. The results in a 1.0 value.

### LINKED ISSUE #926: The `load_demo` function should load in datetimes as strings with their original format intact
### Environment Details
* SDMetrics version: 0.30.0
* Python version: 3.12

### Error Description
The `load_demo` function from SDMetrics seems to automatically cast all datetime columns to `datetime64[ns]`, which inherently strips the values from their actual format. As a result, we can't really apply the [DatetimeFormatAdherence](https://docs.sdv.dev/sdmetrics/data-metrics/diagnostic/datetimeformatadherence) metric.

### Steps to reproduce
Any datetime columns in the demo should be loaded in as strings in order to keep the formatting information preserved. (Later on, if any metric/report needs to convert them to a `datetime64` or `<M8[ns]` dtype, it can always use the `datetime_format` information from the metadata.)

```python

real_data, synthetic_data, metadata = load_demo(modality='single_table')
print(real_data['student_placements']['start_date'].dtype)
print(synthetic_data['student_placements']['start_date'].dtype)
```
```
datetime64[ns]
datetime64[ns]
```

This is different from SDV. In SDV, the `load_demo` function is configured to keep these values as strings and so the whole column is present as an `object` dtype (which is what we want).
```python

    modality='single_table',
    dataset_name='student_placements'
)

print(data['start_date'].dtype)
```
```
object
```

sarahmish:
Resolve #925 
Resolve #926

### DIFF
--- a/sdmetrics/demos.py
+++ b/sdmetrics/demos.py
@@ -7,12 +7,7 @@ import pandas as pd
 
 
 def _load_table(metadata, path):
-    datetime_columns = []
-    for column, column_meta in metadata['columns'].items():
-        if column_meta['sdtype'] == 'datetime':
-            datetime_columns.append(column)
-
-    return pd.read_csv(path, parse_dates=datetime_columns)
+    return pd.read_csv(path)
 
 
 def load_demo(modality='multi_table'):
--- a/sdmetrics/demos.py
+++ b/sdmetrics/demos.py
@@ -6,9 +6,6 @@ import pathlib
 import pandas as pd
 
 
-def _load_table(metadata, path):
-    return pd.read_csv(path)
-
 
 def load_demo(modality='multi_table'):
     """Load demo data of the indicated data modality.
@@ -33,8 +30,8 @@ def load_demo(modality='multi_table'):
     real_data = {}
     synthetic_data = {}
     for table, table_meta in metadata['tables'].items():
-        real_data[table] = _load_table(table_meta, demo_path / f'{table}_real.csv')
-        synthetic_data[table] = _load_table(table_meta, demo_path / f'{table}_synthetic.csv')
+        real_data[table] = pd.read_csv(demo_path / f'{table}_real.csv')
+        synthetic_data[table] = pd.read_csv(demo_path / f'{table}_synthetic.csv')
 
     return real_data, synthetic_data, metadata
 
--- a/sdmetrics/demos.py
+++ b/sdmetrics/demos.py
@@ -6,7 +6,6 @@ import pathlib
 import pandas as pd
 
 
-
 def load_demo(modality='multi_table'):
     """Load demo data of the indicated data modality.
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -8,13 +8,7 @@ from sdmetrics.reports.base_report import BaseReport
 
 @pytest.fixture
 def converted_datetime_single_table_demo():
-    """Single table demo data with the datetime columns converted to ``datetime64``.
-
-    The demo data stores the datetime columns in their original string representation.
-    Metrics that work on the values of a column, such as ``BoundaryAdherence``, need them
-    as ``datetime64``, which is the conversion the reports do before running any property.
-    Use this fixture for the tests that run a property or a metric without a report.
-    """
+    """Single table demo data with the datetime columns converted to ``datetime64``."""
     real_data, synthetic_data, metadata = load_demo(modality='single_table')
     table_name = 'student_placements'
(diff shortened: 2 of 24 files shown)


sdmetrics/demos.py
```diff
@@ -7,12 +7,7 @@
 
 
 def _load_table(metadata, path):
-    datetime_columns = []
-    for column, column_meta in metadata['columns'].items():
-        if column_meta['sdtype'] == 'datetime':
-            datetime_columns.append(column)
-
-    return pd.read_csv(path, parse_dates=datetime_columns)
+    return pd.read_csv(path)
```
R-Palazzo, line 10: Could we remove the metadata from the method's inputs?

tests/integration/conftest.py
```diff
@@ -0,0 +1,42 @@
+"""Fixtures shared by the integration tests."""
+
+import pytest
+
+from sdmetrics.demos import load_demo, load_timeseries_demo
+from sdmetrics.reports.base_report import BaseReport
+
+
+@pytest.fixture
+def converted_datetime_single_table_demo():
```
R-Palazzo, line 10: Could we name it `data_single_table_demo` instead? Since it’s used in many places that don’t specifically focus on datetime conversion.

tests/integration/reports/multi_table/test_diagnostic_report.py
```diff
@@ -209,7 +216,7 @@ def test_end_to_end_with_metrics_failing(self):
                 1.0,
                 1.0,
                 np.nan,
-                1.0,
+                np.nan,
```
R-Palazzo, line 219: That is the change from #925?

R-Palazzo (commented):
Looks good

I just left minor comments

tests/integration/conftest.py
sarahmish: I can do that but I kept it verbose to make it clearer.

For example, boundary, column shapes, coverage, etc tests fail if the datetime column is not in a datetime64 type. This makes it easy to follow what's the data that's being used in the test.

Let me know if you still prefer this change

tests/integration/reports/multi_table/test_diagnostic_report.py
sarahmish: yep! `DatetimeFormatAdherence` should be NaN if the column is in datetime64 format

tests/integration/conftest.py
frances-h: Instead of `converted` could we be explicit and say `object_datetime` or `string_datetime`?

frances-h (approved):

R-Palazzo (approved):

PR #936, opened 2026-09-10
Add constraint validity property

### LINKED ISSUE #900: Add `ConstraintValidity` property
### Problem Description
_As a user, I would like a property that computes whether my data follows the defined constraints._

Once we have created the ConstraintAdherence (https://github.com/sdv-dev/SDMetrics/issues/899), we'd like to create a new property `ConstraintValidity` that checks if each row in the synthetic data is valid according to the defined constraints.

### Expected behavior
- Add `ConstraintValidity` property. This will be added to the Diagnostic report in #901 .

#### Score Calculation
The average of all the `ConstraintAdherence` scores becomes the overall Constraint Validity property score.
* If there are constraints, then we should show a progress bar that increments 1 time for each constraint.
* If there are no constraints, then the property score is NaN

#### Methods
- `_generate_details(self, real_data, synthetic_data, metadata, constraints, progress_bar=None)`: This method will generate the `details` dataframe for the property. 
        -  `Constraint`: The constraint class name that the metric is being calculated for
        - `Metric`: The metric that is applied to the given column
        - `Parameters`: The parameters defined by the constraint
    - The progress bar should update for after each constraint has been evaluated.
- `get_visualization(self)`: There is no visualization for this property yet. It should show the user a friendly error.
    - "Error: No visualization is available for Constraint Validity. To see the detailed score breakdowns, use the 'get_details' function."

sarahmish:
Resolve #900


sdmetrics/reports/multi_table/_properties/constraint_validity.py
```diff
@@ -0,0 +1,122 @@
...
+
+        self.details = pd.DataFrame({
+            'Constraint': constraint_names,
+            'Metric': [ConstraintAdherence.__name__] * len(constraint_names),
+            'Parameters': constraint_parameters,
+            'Score': scores,
+            'Error': error_messages,
+        })
+
+    def get_score(self, real_data, synthetic_data, metadata, constraints=None, progress_bar=None):
```
sarahmish, line 69: overriding this method to have `constraints`

sdmetrics/reports/multi_table/_properties/constraint_validity.py
```diff
@@ -0,0 +1,122 @@
...
+
+        Raises:
+            VisualizationUnavailableError
+        """
+        raise VisualizationUnavailableError(
+            'Error: No visualization is available for Constraint Validity. To see the '
+            "detailed score breakdowns, use the 'get_details' function."
+        )
+
+    def get_details(self, table_name=None):
```
sarahmish, line 111: overriding `get_details` to always return the details dataframe since this property is not defined by tables.

tests/unit/reports/multi_table/_properties/test_constraint_validity.py
```diff
@@ -0,0 +1,304 @@
...
+        # Setup
+        constraint_validity = ConstraintValidity()
+
+        # Assert
+        assert isinstance(constraint_validity, BaseMultiTableProperty)
+        assert constraint_validity._num_iteration_case == 'constraint'
+        assert constraint_validity.is_computed is False
+        assert constraint_validity.details.empty
+
+    def test__get_num_iterations(self, metadata, constraints):
```
R-Palazzo, line 86: We could parameterize this test

tests/unit/reports/multi_table/_properties/test_constraint_validity.py
```diff
@@ -0,0 +1,304 @@
...
+            'Constraint': ['FixedCombinations'],
+            'Metric': ['ConstraintAdherence'],
+            'Parameters': [{'table_name': 'sessions', 'column_names': ['device', 'os']}],
+            'Score': [1.0],
+        })
+
+        # Run
+        details = constraint_validity.get_details()
+
+        # Assert
```
R-Palazzo, line 286: We could add:
```python
assert details is not constraint_validity.details
```

sdmetrics/reports/multi_table/_properties/constraint_validity.py
```diff
@@ -0,0 +1,122 @@
...
+                score = ConstraintAdherence.compute(real_data, synthetic_data, metadata, constraint)
+                error_message = None
+            except Exception as e:
+                score = np.nan
+                error_message = f'{type(e).__name__}: {e}'
+            finally:
+                if progress_bar:
+                    progress_bar.update()
+
+            is_dict = isinstance(constraint, dict)
```
R-Palazzo, line 55: Can `constraint` be anything other than a `dict`?

R-Palazzo (commented):
Looking good!

Thanks for the tests, they cover the different cases well.

I just have one question about constraints inside the `constraints` list that are not dictionaries. Is this something we want to support?
Should we validate that the `constraints` parameter is either `None` or a list of dictionaries?

sarahmish:
> Looking good!
> 
> Thanks for the tests, they cover the different cases well.
> 
> I just have one question about constraints inside the `constraints` list that are not dictionaries. Is this something we want to support? Should we validate that the `constraints` parameter is either `None` or a list of dictionaries?

I added a validation to make sure its either (1) a list of dictionaries (2) an empty list (3) None

sdmetrics/reports/multi_table/_properties/constraint_validity.py
sarahmish: after applying the validation, it's no longer necessary to make sure it's a dict

R-Palazzo (approved):

frances-h (approved):

